PR #17, opened 2017-11-29
[gui] implemented camera page

spenfarr:


base_station/gui/src/cam.html
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    camera {
+      height: 100%;
+    }
+  </style>
+</head>
+
+<body>
+  <camera></camera>
```
moosingin3space: `<camera>` is invalid HTML. I used `<main>` in index.html because that is a valid HTML element.

PR #29, opened 2018-01-28
[hwi] fix encoder messagge and add set demand

MiloATH:
fixes init of Rover in onboard/talon and add ability to set demand


rayfok:
Superceded by other commit.

PR #36, opened 2018-02-11
[hwi] Add ability to set limit switches

MiloATH:


onboard/talon/src/lowlevel.py
```diff
@@ -47,6 +47,7 @@ def __init__(self, bus, device_id):
                     data=[0 for i in range(0, 8)])
         self.control_5_task = can.send_periodic(
             can.rc['channel'], self.msg_control_5, 0.01)
+        self.set_overide_limit_switches(7)
```
moosingin3space: This is a bit of a magic number, can we use an `Enum` somewhere to abstract this away?

PR #37, opened 2018-02-17
[nav] Defined messages AutonState and NavStatus

prabid:


rover_msgs/AutonState.lcm
```diff
@@ -0,0 +1,5 @@
+package rover_msgs;
+
+struct Auton {
```
moosingin3space: Rename `Auton` to `AutonState` in the struct name.

PR #40, opened 2018-02-25
[teleop] sa controls [gui] speed limiter [ansible] pis

spenfarr:


ansible/roles/pis/files/pi-heartbeat.service
```diff
@@ -0,0 +1,8 @@
+[Unit]
+Description=Pi Heartbeat
+After=systemd-modules-load.service
```
moosingin3space: It may be necessary to do an `After=network-online.target` to ensure the network is up before the heartbeater runs.

ansible/roles/pis/tasks/main.yml
```diff
@@ -3,15 +3,60 @@
...
+      src: "files/{{ item }}.service"
+      dest: "/etc/systemd/system/{{ item }}.service"
+  with_items:
+      - uv4l
+      - pi-heartbeat
+
+- name: Disable Mbed compilation
+  copy:
+      src: files/mrover.site
+      dest: /home/mrover.site
```
moosingin3space: Should probably be `/home/{{ user }}/mrover.site`

moosingin3space:
`systemd` service files need to have an `[Install]` section, which specifies when they should run, as not all services actually start on boot. What you probably want is:
```
[Install]
WantedBy=multi-user.target
```
at the end of both of your unit files. Also make sure that the processes are launched as the correct user.

ansible/roles/pis/tasks/main.yml
```diff
@@ -3,15 +3,60 @@
...
+      src: "files/{{ item }}.service"
+      dest: "/etc/systemd/system/{{ item }}.service"
+  with_items:
+      - uv4l
+      - pi-heartbeat
+
+- name: Disable Mbed compilation
+  copy:
+      src: files/mrover.site
+      dest: /home/{{ user }}/mrover.site
```
moosingin3space: Your user variable name is actually `running_user`, not `user`.

moosingin3space:
Closed in favor of https://github.com/umrover/mrover-workspace/commit/0ba6b52cd2a27a332f8a843a1e72bcb64a5fd532.

PR #65, opened 2018-04-22
[hwi] Add reading serial

MiloATH:


moosingin3space:
What is this pull request for? Did we not merge the reader for framed-serial previously?

MiloATH:
We merged the work by Justin for the microcontroller and sending framed-serial. I thought it would be a good idea if someone that wasn't me did code review/merging. This is the code for reading framed-serial.

PR #73, opened 2018-05-14
[teleop] cleanup teleop & add map scale

spenfarr:


moosingin3space:
I think HWI still publishes `Encoder` messages, so don't delete the message type just yet.

PR #74, opened 2018-05-25
[gstreamer] added resolution switching

spenfarr:


pi_zeroes/video/src/__main__.py
```diff
@@ -34,9 +38,11 @@ def start_pipeline():
     fd = vid_process.stdout.fileno()
 
     pipeline_string = ("fdsrc fd=" + str(fd) +
-                       " ! video/x-h264, width=1280, height=720"
+                       " ! video/x-h264, width=" + height +
+                       ", height=" + width +
```
moosingin3space: Why are the width and height inverted here?

moosingin3space:
This PR is the last one before code freeze. Once you've made the changes I requested, squash into a single commit, then I'll merge.

jtwhit:
I don't like how we have two sinks on the base station here. I think that only one should be necessary, but you would need to test it. I don't know if you can do that before the code freeze.

PR #81, opened 2018-10-03
[gui] Implementation of gui_webpack moved into master branch

adamcbrown:
Still needs to be tested with the actual rover


MiloATH:
👍 Would've preferred a rebase.

PR #92, opened 2018-11-06
[talon] Listen for arm demand LCM messages

jubeemer:


onboard/talon/src/__main__.py
```diff
@@ -188,6 +188,6 @@ def main():
     lcm_.subscribe("/setparam", set_param_callback)
     lcm_.subscribe("/setdemand", set_demand_callback)
```
jtwhit: This needs to be changed to set_demand.

PR #121, opened 2019-02-07
[cv] added readme and simplified build variables

jeqyin:
[cv] added readme

[cv] got rid of build garbage.


jtwhit:
We should do this properly, but I need to update jarvis to do so. I'll merge this for now.

PR #134, opened 2019-02-16
[hwi] RGB sensors with multiplexer

chenandr:


beaglebone/rgb/src/__main__.py
```diff
@@ -0,0 +1,39 @@
+# main function for rgb sensor and i2c multiplexer
+from i2c_multiplexer import tca_select
```
jtwhit: This should be `from .i2c_multiplexer import tca_select`.

beaglebone/rgb/src/__main__.py
```diff
@@ -0,0 +1,39 @@
+# main function for rgb sensor and i2c multiplexer
+from i2c_multiplexer import tca_select
+import rgb_sensor
```
jtwhit: This should be `from . import rgb_sensor`.

PR #167, opened 2019-04-07
[nav-filter] implement first version of sensor fusion algorithm

amverni:
Caution: LocationFilter1 is still under construction

### DIFF
--- a/onboard/filter/src/filterObjects.py
+++ b/onboard/filter/src/filterObjects.py
@@ -136,10 +136,23 @@ class LocationEstimate:
 
     def derive_location(self):
         """Fix location data after calculations."""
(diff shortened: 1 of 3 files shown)


onboard/filter/src/filterObjects.py
```diff
@@ -136,11 +136,24 @@ def check_validity(self):
 
     def derive_location(self):
```
jtwhit: This function is way overcomplicated. The same can be implemented with simply
```python
temp_lat = self._lat_deg
self._lat_deg = math.trunc(temp_lat)
self._lat_min = (temp_lat - self._lat_deg) * 60

temp_long = self._long_deg
self._long_deg = math.trunc(temp_long)
self._long_deg = (temp_long - self._long_deg) * 60
```

PR #169, opened 2019-04-10
[nav] Integraded abstracted obstacle to upstream. Ready for pull request

elgarber:
Abstracted obstacle avoidance code


onboard/nav/avoidances.cpp
```diff
@@ -0,0 +1,164 @@
...
+        case AvoidanceType::UNKNOWN:
+            std::cerr << "Unkown Search Type. Defaulting to original\n";
+            avoid = new Original( stateMachine );
+            break;
+    }
+    return avoid;
+}
+
+/*************************************************************************/
+/* Original Obstacle Avoidance Algorithm
```
amverni: Again I suggest using a better name than `Original` in favor of something more descriptive. If nothing strikes you my suggestion is `Simple` since we are performing the most simple of avoidance algorithms we can by inserting one point into our path to avoid create a new path to avoid the obstacle. I think the comment should also include something similar to this so as to describe how this obstacle avoidance algorithm solves our problem.

onboard/nav/avoidances.cpp
```diff
@@ -0,0 +1,164 @@
...
+// If the rover is turned off, it proceeds to Off.
+// If the tennis ball is detected, the rover proceeds to it.
+// If the obstacle is no longer detcted, proceed to drive around the obstacle.
+// Else continue turning around the obstacle.
+// ASSUMPTION: We assume that the obstacle is straight ahead of us,
+//             therefore we produce an underestimate for how close the waypoint is to the
+//             obstacle. This relies on using a path width no larger than what we can
+//             confidentally see to the side.
+//
+//             There is no rock that is more than 8 meters in diameter
```
amverni: `// ASSUMPTION: ...` to make it clear this is also an assumption and is a separate assumption. This way you can also remove the blank line above

onboard/nav/avoidances.cpp
```diff
@@ -0,0 +1,164 @@
...
+NavState Original::executeTurnAroundObs( Rover * mPhoebe, const rapidjson::Document& roverConfig )
+{
+
+    double cvThresh = roverConfig[ "computerVision" ][ "obstacleThresh" ].GetDouble();
```
amverni: We can now use the depth to the obstacle sent over by CV

onboard/nav/avoidances.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    //if we see the ball and are in a search state, only go to turn to ball if turning
```
amverni: This comment is missing some words and doesn't quite make sense
Minor: `// If`

onboard/nav/avoidances.cpp
```diff
@@ -0,0 +1,164 @@
...
+    //   same direction for obstcle
+    if( mPhoebe->roverStatus().currentState() == NavState::SearchTurnAroundObs &&
+        mPhoebe->roverStatus().tennisBall().found &&
+      ( cvThresh > mPhoebe->roverStatus().tennisBall().distance - 2 or  //TODO Replace cvThresh with dist to obs
```
amverni: Do this `TODO` and maybe abstract part or all of this if statement into a function because it is a lot for a single if statement. Maybe make a function that will return true if you're in a search state and see the ball and then combine the other lines into a logical function so you only have two things check in the if.

onboard/nav/avoidances.cpp
```diff
@@ -0,0 +1,164 @@
...
+      ( mPhoebe->roverStatus().tennisBall().bearing > 0 && mPhoebe->roverStatus().obstacle().bearing > 0 ) ) )
+    {
+        return NavState::TurnToBall;
+    }
+
+    if( !mPhoebe->roverStatus().obstacle().detected )
+    {
+        double distanceAroundObs = cvThresh / cos( fabs( degreeToRadian( mOriginalObstacleAngle ) ) );
+        mObstacleAvoidancePoint = createAvoidancePoint( mPhoebe, distanceAroundObs ) ;
+
```
amverni: remove this whitespace

onboard/nav/avoidances.cpp
```diff
@@ -0,0 +1,164 @@
...
+    //   same direction for obstcle
+    if( mPhoebe->roverStatus().currentState() == NavState::SearchTurnAroundObs &&
+        mPhoebe->roverStatus().tennisBall().found &&
+      ( cvThresh > mPhoebe->roverStatus().tennisBall().distance - 2 or  //TODO Replace cvThresh with dist to obs
+      ( mPhoebe->roverStatus().tennisBall().bearing < 0 && mPhoebe->roverStatus().obstacle().bearing < 0 ) or
+      ( mPhoebe->roverStatus().tennisBall().bearing > 0 && mPhoebe->roverStatus().obstacle().bearing > 0 ) ) )
+    {
+        return NavState::TurnToBall;
+    }
+
```
amverni: Minor: I'd appreciate if we remove this blank line as it is not needed for readability since the if statements are already separating it. Please do this throughout as below there is some more. Moreover, the two blank lines a few lines down is too much as it makes it seem that the two blocks of code are quite separated when they actually are quite dependent. (This doesn't mean you have to remove all whitespace but we don't want to overuse it as it makes the code slower to skim and we can't look at as much at once.)

onboard/nav/avoidances.cpp
```diff
@@ -0,0 +1,164 @@
...
+// If the tennis ball is detected, the rover proceeds to it.
+// If the obstacle is no longer detcted, proceed to drive around the obstacle.
+// Else continue turning around the obstacle.
+// ASSUMPTION: We assume that the obstacle is straight ahead of us,
+//             therefore we produce an underestimate for how close the waypoint is to the
+//             obstacle. This relies on using a path width no larger than what we can
+//             confidentally see to the side.
+//
+//             There is no rock that is more than 8 meters in diameter
+NavState Original::executeTurnAroundObs( Rover * mPhoebe, const rapidjson::Document& roverConfig )
```
amverni: Make `mPhoebe` into `phoebe` because it is not a member. Please do this throughout the files and be consistent with the placement of `*` in `Rover* pheobe`. While I'd prefer `Rover* pheobe` to be consistent with the rest of our code, whichever you choose doesn't matter but it should be consistent inside the file at least

onboard/nav/avoidances.cpp
```diff
@@ -0,0 +1,164 @@
...
+            return NavState::DriveAroundObs;
+        }
+        mJustDetectedObstacle = false;
+
+        return NavState::SearchDriveAroundObs;
+    }
+
+    double obstacleBearing = mPhoebe->roverStatus().obstacle().bearing;
+
+    if (mJustDetectedObstacle && (obstacleBearing < 0) ? (mLastObstacleAngle >= 0) : (mLastObstacleAngle < 0)) {
```
amverni: Do we know for sure the order of precedence here? For clarity can we make this `mJustDetectedObstacle` && (obstacleBearing < 0 ? mLastObstacleAngle >= 0 : mLastObstacleAngle < 0)`. I could honestly be wrong and what you are trying to do here is `(mJustDetectedObstacle` && obstacleBearing) < 0 ? mLastObstacleAngle >= 0 : mLastObstacleAngle < 0` but that's why this change is necessary

onboard/nav/avoidances.cpp
```diff
@@ -0,0 +1,164 @@
...
+        {
+            return NavState::DriveAroundObs;
+        }
+        mJustDetectedObstacle = false;
+
+        return NavState::SearchDriveAroundObs;
+    }
+
+    double obstacleBearing = mPhoebe->roverStatus().obstacle().bearing;
+
```
amverni: please put a concise comment as to why you are multiplying by one in this case (and maybe say what **this** case is which might even be enough to explain the why) as it is not inherently clear

onboard/nav/avoidances.cpp
```diff
@@ -0,0 +1,164 @@
...
+    if (mJustDetectedObstacle && (obstacleBearing < 0) ? (mLastObstacleAngle >= 0) : (mLastObstacleAngle < 0)) {
+        obstacleBearing *= -1;
+    }
+
+    double desiredBearing = mod( mPhoebe->roverStatus().odometry().bearing_deg
```
amverni: try not to split the first argument onto two lines unless it really helps readability

onboard/nav/avoidances.cpp
```diff
@@ -0,0 +1,164 @@
...
+                               + obstacleBearing, 360 );
+                           //std::cout << "\nobstacle bearing: " << mPhoebe->roverStatus().obstacle().bearing << std::endl;
+
+    mJustDetectedObstacle = true;
+    mLastObstacleAngle = obstacleBearing;
+
+    mPhoebe->turn( desiredBearing );
+    // std::cout << "here" << mPhoebe->roverStatus().odometry().bearing_deg << " " << obstacleBearing << " " << desiredBearing << std::endl;
```
amverni: remove

onboard/nav/avoidances.cpp
```diff
@@ -0,0 +1,164 @@
...
+    return mPhoebe->roverStatus().currentState();
+} // executeTurnAroundObs()
+
+// Executes the logic for driving around an obstacle.
```
amverni: If possible, make this comment more concise. It doesn't need to lay out every detail of the function.

onboard/nav/avoidances.cpp
```diff
@@ -0,0 +1,164 @@
...
+        1. the difference in distance (tennis ball is closer than obs)
+        2. Turning around obs is same direction to turn to tennis ball */
+/*************************************************************************/
+Original::~Original() {}
+
+/*****************************************************/
+/* Helpers */
+/*****************************************************/
+
+// Executes the logic for turning around an obstacle.
```
amverni: If possible, make this comment more concise. It doesn't need to lay out every detail of the function. The assumption does not need to be condensed as it is very important to convey that message.

onboard/nav/avoidances.cpp
```diff
@@ -0,0 +1,164 @@
...
+} // executeTurnAroundObs()
+
+// Executes the logic for driving around an obstacle.
+// If the rover is turned off, proceed to Off.
+// If another obstacle is detected, proceed to go around it.
+// If the rover finished going around the obstacle, it proceeds to
+//      turn to the Waypoint that was being driven to when the
+//      obstacle was spotted.
+// Else, continue driving around the obstacle.
+NavState Original::executeDriveAroundObs( Rover * mPhoebe )
```
amverni: `mPheobe` -> `phoebe`

onboard/nav/avoidances.cpp
```diff
@@ -0,0 +1,164 @@
...
+        return mPhoebe->roverStatus().currentState();
+    }
+    if( mPhoebe->roverStatus().currentState() == NavState::DriveAroundObs )
+    {
+        return NavState::TurnAroundObs;
+    }
+    return NavState::SearchTurnAroundObs;
+} // executeDriveAroundObs()
+
+// Creates the odometry point used to drive around an obstacle
```
amverni: Minor: `// Create ...`

onboard/nav/avoidances.cpp
```diff
@@ -0,0 +1,164 @@
...
+Odometry Original::createAvoidancePoint( Rover * mPhoebe, const double distance )
```
amverni: `mPheobe` -> `phoebe`

onboard/nav/avoidances.cpp
```diff
@@ -0,0 +1,164 @@
...
+    double totalLatitudeMinutes = avoidancePoint.latitude_min +
+        cos( degreeToRadian( avoidancePoint.bearing_deg ) ) * distance * LAT_METER_IN_MINUTES;
+    double totalLongitudeMinutes = avoidancePoint.longitude_min +
+        sin( degreeToRadian( avoidancePoint.bearing_deg ) ) * distance * mPhoebe->longMeterInMinutes();
+    avoidancePoint.latitude_deg += totalLatitudeMinutes / 60;
+    avoidancePoint.latitude_min = ( totalLatitudeMinutes - ( ( (int) totalLatitudeMinutes) / 60 ) * 60 );
+    avoidancePoint.longitude_deg += totalLongitudeMinutes / 60;
+    avoidancePoint.longitude_min = ( totalLongitudeMinutes - ( ( (int) totalLongitudeMinutes) / 60 ) * 60 );
+
+    // std::cout << distance << std::endl;
```
amverni: remove these commented out debug statements

onboard/nav/avoidances.hpp
```diff
@@ -0,0 +1,39 @@
...
+enum class AvoidanceType
+{
+    ORIGINAL,
+    UNKNOWN
+};
+
+Avoidance* AvoidFactory( StateMachine* stateMachine, AvoidanceType type );
+
+/*************************************************************************/
+/* Default Avoidance */
```
amverni: This comment doesn't seem super helpful nor accurate. It is currently our only avoidance algorithm not our default.

onboard/nav/avoidances.hpp
```diff
@@ -0,0 +1,39 @@
...
+/*************************************************************************/
+class Original : public Avoidance
```
amverni: Add comment for the class. This should be in the form `// This class ...` rather than `/*****/ comment /******/`

onboard/nav/avoidances.hpp
```diff
@@ -0,0 +1,39 @@
...
+{
+public:
+    Original( StateMachine* stateMachine_ ) 
+    : Avoidance(stateMachine_) {}
+
+    NavState executeTurnAroundObs( Rover* phoebe, const rapidjson::Document& roverConfig );
+
+    NavState executeDriveAroundObs( Rover* phoebe );
+
+    Odometry createAvoidancePoint( Rover * phoebe, const double distance );
```
amverni: Consistency with `*` placement

onboard/nav/avoidances.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+    ~Original();
```
amverni: Put the destructor under the constructor

onboard/nav/avoidances.hpp
```diff
@@ -0,0 +1,39 @@
...
+    
+};
+
+#endif //AVOIDANCES_HPP
+
+
+/*************************************************************************/
```
amverni: Remove if you have no `TODO`s

onboard/nav/searcher.cpp
```diff
@@ -13,11 +13,11 @@ Odometry Searcher::frontSearchPoint( )
     return mSearchPoints.front();
 }
 
-void Searcher::popSearchPoint( )
-{
-    mSearchPoints.pop_front();
-    return;
-}
+// void Searcher::popSearchPoint( )
```
amverni: Remove this function if it's not necessary

onboard/nav/searcher.cpp
```diff
@@ -231,22 +231,29 @@ NavState Searcher::executeTurnToBall( Rover * phoebe )
...
+NavState Searcher::executeDriveToBall( Rover * phoebe, const rapidjson::Document& roverConfig )
 {
     if( !phoebe->roverStatus().tennisBall().found )
     {
         cerr << "Lost the tennis ball\n";
         return NavState::SearchFaceNorth;
     }
-    if( phoebe->roverStatus().obstacle().detected )
+
+    double cvThresh = roverConfig[ "computerVision" ][ "obstacleThresh" ].GetDouble();
```
amverni: Use distance from CV message unless it hurts our performance

onboard/nav/searcher.cpp
```diff
@@ -231,22 +231,29 @@ NavState Searcher::executeTurnToBall( Rover * phoebe )
...
+    
+    if( phoebe->roverStatus().obstacle().detected and
```
amverni: This if statement can be made more readable. Probably through the use of functions. In the function explain why we are subtracting 2 as well since this is not intuitive. My only guess is for being within 2 meters of a TB but even this is a guess. Also, we now attempt to get within 1 meter of TBs since 2.01 meters is 0 points.

onboard/nav/searcher.hpp
```diff
@@ -19,7 +19,7 @@ class Searcher {
 
     Odometry frontSearchPoint();
 
-    void popSearchPoint();
+    // void popSearchPoint();
```
amverni: Remove if not necessary

onboard/nav/stateMachine.hpp
```diff
@@ -121,6 +116,9 @@ class StateMachine
     // Search pointer to control search states
     Searcher* mSearcher;
 
+    // Avoidance pointer to control obstacle avoidance states
+    Avoidance* mObstacle;
```
amverni: This comment and variable name don't seem to explain what this does. Maybe `mObstacleAvoidanceStateMachine` or something similar

onboard/nav/avoidance.hpp
```diff
@@ -0,0 +1,54 @@
...
+    void updateObstacleAngle( double bearing );
+
+    NavState run( Rover * phoebe, const rapidjson::Document& mRoverConfig );
+
+    virtual Odometry createAvoidancePoint( Rover * phoebe, const double distance ) = 0;
+
+    virtual NavState executeTurnAroundObs( Rover* phoebe, const rapidjson::Document& mRoverConfig ) = 0;
+
+    virtual NavState executeDriveAroundObs( Rover* phoebe ) = 0;
+
```
amverni: minor: Lots of unneeded whitespace

onboard/nav/avoidances.cpp
```diff
@@ -0,0 +1,164 @@
+/* Searches.cpp */
```
amverni: Be careful copy and pasting code.
Also, please make this file only for the one obstacle avoidance algorithm. We will make more files when we make a new algorithm. This will help with conflicts in development. Ask David if you don't believe me as he suffered before I learned my lesson here.

Otherwise the logic looks generally good for the PR.

onboard/nav/avoidance.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+
+protected:  // TODO
+    /*************************************************************************/
+    /* Protected Member Variables */
+    /*************************************************************************/
+    StateMachine* stateMachine;
```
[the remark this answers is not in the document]
elgarber: I changed the name to roverStateMachine because it's not really a member variable. I don't want it to get confused with the other pointer. (Also, I think this is what you asked above when changing from stateMachine( stateMachine_ ) to roverStateMachine( stateMachine_ ) in the constructor.

onboard/nav/avoidances.hpp
```diff
@@ -0,0 +1,39 @@
+#ifndef AVOIDANCES_HPP
+#define AVOIDANCES_HPP
+
+#include "avoidance.hpp"
+
+// This class is the representation of different avoidance algorithms
+enum class AvoidanceType
+{
+    ORIGINAL,
+    UNKNOWN
```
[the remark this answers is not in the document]
elgarber: We have it in the NavState enum class, so I was following that. But I'll remove it

onboard/nav/avoidance.hpp
elgarber: I thought this was the style we were using. Should there be no space between functions in a .hpp file class declaration? Isn't this how it is in every file? The spaces I have are the same as StateMachine.hpp

onboard/nav/avoidance.cpp
```diff
@@ -0,0 +1,42 @@
+#include "avoidance.hpp"
```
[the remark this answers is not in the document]
elgarber: The reason I put all algorithms in one file is because if we make them all their own files, we could end up including a lot of files in the header

onboard/nav/obstacleAvoidanceStateMachine.hpp
```diff
@@ -0,0 +1,71 @@
...
+// ObstacleAvoidanceFactory creates an ObstacleAvoidanceStateMachine object based on
+// the inputted obstacle avoidance algorithm. This allows for an an ease of transition 
+// between obstacle avoidance algorithms
+ObstacleAvoidanceStateMachine* ObstacleAvoiderFactory( StateMachine* roverStateMachine, 
+                                                       ObstacleAvoidanceAlgorithm algorithm );
+
+#endif //OBSTACLEAVOIDANCESTATEMACHINE_HPP
+
+// TODO: 
+//  remove protected. Bad practice to use protected ( they could simply be public instead )
```
amverni: this should not be a `TODO`. If this **needs** to be fixed, then now is the time to fix it. I feel that it does not need to be changed and the protected is okay here since only derived classes will use these members so there is no need to make it public and only MRover will be using this so we don't need to make getters/setters

onboard/nav/searcher.cpp
```diff
@@ -13,16 +13,10 @@ Odometry Searcher::frontSearchPoint( )
...
 
-// void Searcher::popSearchPoint( )
-// {
-//     mSearchPoints.pop_front();
-//     return;
-// }
-
 /*****************************************************/
 /* Searcher Run Fuction */
 /*****************************************************/
```
amverni: (see last review) fix this comment

onboard/nav/avoidance.cpp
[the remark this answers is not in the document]
amverni: Have a few extra header files is not a problem. I would prefer this for the benefit of added readability

onboard/nav/obstacleAvoidanceStateMachine.cpp
```diff
@@ -0,0 +1,84 @@
...
+bool ObstacleAvoidanceStateMachine::isTennisBallDetected ( Rover* phoebe ) 
+{
+    return ( phoebe->roverStatus().currentState() == NavState::SearchTurnAroundObs &&
+             phoebe->roverStatus().tennisBall().found );
+}
+
+// Checks to see if tennis ball is reachable before hitting obstacle
+// Tennis ball must be closer than obstacle and must be in same direction ( in other words, 
+// the rover should be turning the same way to go around the obstacle as to get to the ball )
+bool ObstacleAvoidanceStateMachine::tennisBallReachable( Rover* phoebe, double distance,
```
amverni: add verb to function name (I.e. `isTennisBallReachable`)

onboard/nav/searcher.cpp
```diff
@@ -268,6 +259,13 @@ NavState Searcher::executeDriveToBall( Rover * phoebe, const rapidjson::Document
     return NavState::TurnToBall;
 } // executeDriveToBall()
 
+bool Searcher::tennisBallReachable( Rover* phoebe, double distance, double bearing )
```
amverni: This function is identical to one defined in `obstacleAvoidanceStateMachine.cpp`. We should put it somewhere that both these files have access to (e.g. `utilities.cpp`, `rover.cpp`, etc., wherever makes the most sense)

onboard/nav/searcher.hpp
```diff
@@ -11,16 +11,16 @@ class Searcher {
     /* Public Member Functions */
     /*************************************************************************/
     Searcher(StateMachine* stateMachine_)
-    : stateMachine(stateMachine_) {}
+    : roverStateMachine(stateMachine_) {}
```
amverni: function definitions should be in the cpp file

onboard/nav/searches.cpp
```diff
@@ -164,6 +164,10 @@ void LawnMower::initializeSearch( Rover* phoebe, const rapidjson::Document& rove
             nextSearchPoint.longitude_deg += totalLongitudeMinutes / 60;
             nextSearchPoint.longitude_min = ( totalLongitudeMinutes - ( ( (int) totalLongitudeMinutes) / 60 ) * 60 );
 
+
```
amverni: I believe this was from debugging search with Ana, let's remove it since we finished that and likely won't need it in the same way again.

onboard/nav/obstacleAvoidanceStateMachine.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+        case NavState::DriveAroundObs:
+        case NavState::SearchDriveAroundObs:
+        {
+            return executeDriveAroundObs( phoebe );
+        }
+
+        default:
+        {
+            return NavState::Unknown;
```
amverni: put a `cerr` statement here to help us track down this bug if it were to ever happen

onboard/nav/obstacleAvoidanceStateMachine.cpp
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// Checks to see if tennis ball is reachable before hitting obstacle
+// Tennis ball must be closer than obstacle and must be in same direction ( in other words, 
+// the rover should be turning the same way to go around the obstacle as to get to the ball )
+bool ObstacleAvoidanceStateMachine::tennisBallReachable( Rover* phoebe, double distance, 
+                                                         double bearing ) 
+{
+    return ( ( phoebe->roverStatus().obstacle().distance > distance - 2 ) ||
+     ( bearing < 0 && phoebe->roverStatus().obstacle().bearing < 0 ) ||
```
amverni: Here and throughout, indent to make starting parentheses match

onboard/nav/stateMachine.cpp
```diff
@@ -392,10 +386,16 @@ double StateMachine::getOptimalAvoidanceAngle() const
...
+     ( ( bearing - mPhoebe->roverStatus().odometry().bearing_deg ) > 0 
+                        && mPhoebe->roverStatus().obstacle().bearing < 0 ) );
+}
 
 // thresholds based on state? waypoint vs ball
-// set distance to go around obstacles?
-// set threshold for when to skip a point?
 
 // TODOS:
 // [drive to ball] obstacle and ball
```
amverni: Didn't you fix this `TODO`?

rover_msgs/Obstacle.lcm
```diff
@@ -3,4 +3,5 @@ package rover_msgs;
 struct Obstacle {
 	boolean detected;
 	double bearing; // from straight ahead	
+	double distance; // from straight ahead
```
amverni: It's concerning that we had to add this considering CV was already planning on sending us a distance, we should make sure this is what they are using.

onboard/nav/avoidances.hpp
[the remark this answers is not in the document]
amverni: We use `NavState::Unknown` but we don't use `AvoidanceType::Unknown`

onboard/nav/avoidance.hpp
amverni: That's not quite what I meant. However, you're right that the other files do it the same way as each other so you can use those as a template.

onboard/nav/obstacleAvoidanceStateMachine.hpp
```diff
@@ -0,0 +1,71 @@
+#ifndef OBSTACLEAVOIDANCESTATEMACHINE_HPP
+#define OBSTACLEAVOIDANCESTATEMACHINE_HPP
+
+#include "rover.hpp"
```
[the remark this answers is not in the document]
elgarber: I tried the forward declaration, but it doesn't work. The obstacle avoidance file needs to have direct access to NavState in order to compile.

onboard/nav/searcher.cpp
elgarber: They're not identical. The greater than, less than signs are different.

onboard/nav/stateMachine.cpp
elgarber: yes. The distance to go around obstacles is the distance CV is sending us and we're not missing any waypoints anymore

onboard/nav/stateMachine.hpp
```diff
@@ -63,22 +68,22 @@ class StateMachine
...
-
     string stringifyNavState() const;
 
     bool isObstacleDetected() const;
 
     double getOptimalAvoidanceAngle() const;
 
+    double getOptimalAvoidanceDistance() const;
+
+    std::tuple<double, double> getOptimalAvoidanceElements() const;
```
amverni: Is there a reason to have all `getOptimalAvoidanceAngle`, `getOptimalAvoidanceDistance`, and `getOptimalAvoidanceElements` if they are all private?

onboard/nav/stateMachine.cpp
```diff
@@ -323,13 +337,17 @@ NavState StateMachine::executeTurn()
...
-    if( isObstacleDetected() )
+    const Waypoint& nextWaypoint = mPhoebe->roverStatus().path().front();
+    double distance = estimateNoneuclid( mPhoebe->roverStatus().odometry(), nextWaypoint.odom );    
+    double bearing = calcBearing( mPhoebe->roverStatus().odometry(), nextWaypoint.odom );
+
+    if( isObstacleDetected() && waypointReachable( distance, bearing ) )
     {
-        mOriginalObstacleAngle = getOptimalAvoidanceAngle();
+        updateObstacleElements( getOptimalAvoidanceAngle(), getOptimalAvoidanceDistance() );
+        // updateObstacleAngle( getOptimalAvoidanceAngle() );
```
amverni: remove these commented out lines

onboard/nav/search/searchStateMachine.hpp
```diff
@@ -1,27 +1,30 @@
...
 
 class StateMachine;
 
-class Searcher {
+enum class SearchType
+{
+    SPIRALOUT,
+    LAWNMOWER,
+    SPIRALIN,
+    UNKNOWN
```
amverni: This is the same as for Avoidance type, why do we need `Unkown`? Where you use it in the code, you should just do `default` rather than `case: Unknown`

onboard/nav/rover.cpp
```diff
@@ -160,9 +160,20 @@ bool Rover::turn( double bearing )
...
     throughZero( bearing, mRoverStatus.odometry().bearing_deg );
-    if( fabs( bearing - mRoverStatus.odometry().bearing_deg ) < mRoverConfig[ "navThresholds" ][ "turningBearing" ].GetDouble() )
+    if (mRoverStatus.currentState() == NavState::TurnAroundObs ||
+        mRoverStatus.currentState() == NavState::SearchTurnAroundObs ) 
     {
-        return true;
+        if( fabs( bearing - mRoverStatus.odometry().bearing_deg ) < 0 )
+        {
+            return true;
+        }
```
amverni: We discussed this in person about a week and a half ago but please follow up if needed. Based on our conversation then, I believe we need to add an `else if` for when the bearing is between 0 and some set angle (probably found in the config file so we aren't hardcoding in values) so that we get stuck not giving the rover enough power indefinitely. This `else if` would need to do something to increase the power sent by the joystick LCM. The simple but inelegant solution is to change the bearing to the "some set angle" if in that range.

onboard/nav/search/searchStateMachine.hpp
```diff
@@ -1,27 +1,30 @@
...
+};
+
+class searchStateMachine {
```
amverni: Comment the class name and comment what the class is for.

onboard/nav/rover.cpp
elgarber: 1. I vaguely remember talking about this, but as of now, it will never go into this statement because it fabs. So I think I just didn't make the new config value. 

2. I have no idea how to do the increase in power from the LCM, which I think I told you on that day. It is just changing the upping the turning effort? How much should we do this by?

onboard/nav/stateMachine.hpp
elgarber: We need these if an obstacle is seen from the search algorithm. If you prefer, I can make them public, but maybe we should leave that for next year because we really need to start testing this code.

onboard/nav/rover.cpp
elgarber: I think I see what you mean, and I will try to put it in the code now

elgarber: Apparently, the ramping problem can be fixed in pid tuning

elgarber: If that's not the error, I honestly have no idea what you're talking about.

elgarber: It would never go into this statement anyway because its the absolute value. So it would never be less than 0

PR #175, opened 2019-04-14
[nav] improve search

aswarner:
added search spins throughout search, decreased tennis ball threshold to one meter


onboard/nav/searcher.cpp
```diff
@@ -19,18 +20,48 @@ void Searcher::popSearchPoint( )
...
+            cout << "inserted " << numPoints << " points" << endl;
+            double newDifference = distance / ( numPoints + 1 );
+            double bearing = calcBearing( point1, point2 );
+            cout << "bearing: " << calcBearing( point1, point2 ) << endl;
+            for ( int j = 0; j < numPoints; ++j )
+            {
+                Odometry startPoint = mSearchPoints.at( i );
+                Odometry newOdom = createOdom( startPoint, bearing, newDifference, phoebe );
+                auto iterator = mSearchPoints.begin();
+                for ( int k = 0; k <= i; ++k )
```
amverni: This for loop can be condensed into the line above:
`auto iterator = mSearchPoints.begin() + i + 1;`
If you wanted to get real fancy you could condense lines 46-52 into:
`mSearchPoints.insert(mSearchPoints.begin() + ++i, newOdom);`
I've tested both of these but you should verify that whatever you choose to go with works as well still

onboard/nav/searcher.hpp
```diff
@@ -42,6 +41,13 @@ class Searcher {
     NavState executeDriveToBall( Rover* phoebe );
 
 protected:  // TODO
+    /*************************************************************************/
+    /* Protected Member Functions */
+    /*************************************************************************/
+
+    //function for correcting distances between points
```
amverni: Our function comments typically go in the .cpp files (also start function comments with verb not 'function for ...')

onboard/nav/searcher.hpp
```diff
@@ -42,6 +41,13 @@ class Searcher {
...
+    void correctPoints( Rover * phoebe, const rapidjson::Document& roverConfig );
```
amverni: I'm not sure this function name captures what your function does as you don't actually modify any points. Something more like `insertIntermediatePoints` might be more accurate and help others understand your function better when they see it called.

PR #181, opened 2019-04-24
Add stop commands to force rover to stop

amverni:
Fixes bugs with rover not turning off and not pausing correctly

### DIFF
--- a/onboard/nav/stateMachine.cpp
+++ b/onboard/nav/stateMachine.cpp
@@ -82,7 +82,6 @@ void StateMachine::run()
 {
     if( isRoverReady() )
     {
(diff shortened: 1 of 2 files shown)


onboard/nav/stateMachine.cpp
```diff
@@ -89,7 +89,7 @@ void StateMachine::run()
 
         if( !mPhoebe->roverStatus().autonState().is_auton )
         {
-            nextState = NavState::Off;
+            mPhoebe->roverStatus().currentState() = nextState = executeOff();
```
jtwhit: This is confusing. Split into two lines.

PR #193, opened 2019-04-28
[Jarvis] Revised: Jarvis exec now accepts build dirs

DavidSmith166:
replaces forward slash in first argument to exec with underscores and trims trailing underscores
should no longer interfere with successive arguments 
e.g `jarvis exec lcm_tools_echo Encoder /encoder` will behave as expected.


DavidSmith166:
changes made.

PR #217, opened 2019-05-18
[talon] Added current limiting for drive talons.

jtwhit:


jubeemer:
From the CTRE docs: "For simpler current-limiting (single threshold) use ConfigContinuousCurrentLimit() and set the peak to zero: ConfigPeakCurrentLimit(0)." It seems as though this single limit is the behavior we're going for, so I think the continuous limit should be set to 10 and the peak limit to 0.

PR #235, opened 2019-05-22
[sa] Removed ammonia test automatic servo controls.

jtwhit:
By request of Kitty and Rachel.


jubeemer:
Is the plan to control the servos manually for these tests? How is the post-dip RGB reading going to be taken?

jtwhit:
Yeah. I guess they'll take a reading before and after the dip.

PR #260, opened 2019-12-08
[nav] switch from TennisBall LCM to TargetList

amverni:


amverni:
@calebhar12 Can you look over the CV parts of this patch to ensure we didn't break any of your code?

amverni:
@jubeemer could you double check that the jarvis and teleop commits in here look right. We did not touch those files and to me it seems like it's the exact same code as is already in there so I don't think anything is wrong.

aswarner:
auton code is good, need @calebhar12  to check cv

calebhar12:
cv code looks good!

PR #261, opened 2019-12-08
[nav-sim] Create auton gui page

scsheeha:


amverni:
@prabid could you look this over and make sure there's no duplicated code from things already existing in other parts of the GUI page or any other obvious mistakes from a teleop perspective?

prabid:
My one concern is that you use document.getElementById, which isn't recommended with vue because you are referencing the dom directly, which can lead to performance issues and messier logic. If you can, attach values in the html to Vue variables that you can then reference and modify within the js code.

prabid:
Otherwise, looks good

PR #268, opened 2020-01-21
[LCM] Reworked LCM Messages between HWI and Teleop

raytitan:


jubeemer:
@raytitan What's the deal with the extra commits (merge commits and test bench stuff)?

raytitan:
@jubeemer  do you want me to squash?

jubeemer:
@raytitan For things like IDs and states with a small number of possible options, we should define a mapping to integers and leave a comment with the mapping in the .lcm file. I don't want to make a habit of sending LCMs with unnecessary strings because they're a lot of extra data. Not sure if we're up against a bandwidth issue but we should avoid making them unnecessarily large.

jubeemer:
> @jubeemer do you want me to squash?

Yes, if you want those commits included. Also, I'd prefer not to remove large amounts of code in a commit for something else. If in the future we wanted to go find our RGB program, there should be a commit for it so we can find it.

PR #274, opened 2020-01-28
[nav-sim] Update navigation simulator for radio repeaters and gate directionality

amverni:


simulators/nav/src/constants.js
```diff
@@ -2,3 +2,6 @@ export const WAYPOINT_MODE = 0;
 export const ROCK_MODE = 1;
 export const AR_TAG_MODE = 2;
 export const GATE_MODE = 3;
+export const RADIO_REPEATER_THRESHOLD = 30;
```
amverni: I will update these numbers in a future commit when we know them better. These will work fine for testing in the meantime.

PR #277, opened 2020-01-30
[cv] Integrate AR Tag detection into onboard/cv

y2kmurali:
Now detects AR Tags and pushes them to LCM

Increases OpenCV version dependency from 2.4.9 -> 3.2 due to use of aruco contrib module


onboard/cv/artag_detector.cpp
```diff
@@ -0,0 +1,124 @@
...
+    // and the tag ID number return them such that the "leftmost" (x
+    // coordinate) tag is at index 0
+    cvtColor(src, rgb, COLOR_RGBA2RGB);
+    // clear ids and corners vectors for each detection
+    ids.clear();
+    corners.clear();
+
+#if PERCEPTION_DEBUG
+    // on click debugging for color
+    DEPTH = depth_src;
```
amverni: Why is this being set here and then overwritten two lines later? It doesn't appear to me that cvtColor uses it because it's not an input and cvtColor wouldn't otherwise have access to our variable, right? I'm under the assumption that cvtColor is the openCV function.

onboard/cv/artag_detector.cpp
```diff
@@ -0,0 +1,124 @@
...
+    cvtColor(rgb, HSV, COLOR_RGB2HSV);
+    DEPTH = depth_src;
+    setMouseCallback("image", onMouse);
+#endif
+
+    /// Find tags
+    cv::aruco::detectMarkers(rgb, alvarDict, corners, ids, alvarParams);
+
+#if PERCEPTION_DEBUG
+    // Draw detected tags
```
amverni: Looking at the documentation for `detectMarkers`, rob, corners, and ids are not changed so their values in this section (inside the `#if PERCEPTION_DEBUG`) are the same before and after the call to `detectMarkers`. Therefore, it's my understanding that you can combine this `#if PERCEPTION_DEBUG` section with the one above which will make this section a little more readable.

onboard/cv/artag_detector.cpp
```diff
@@ -0,0 +1,124 @@
...
+    cv::aruco::drawDetectedMarkers(rgb, corners, ids);
+    cv::imshow("AR Tags", rgb);
+#endif
+
+    // create Tag objects for the detected tags and return them
+    pair<Tag, Tag> discoveredTags;
```
amverni: I would suggest adding a default constructor to the Tag class so that you can git rid of a handful of these lines of code. This is not strictly necessary, I just feel it would make your code more readable for others. I think you can also just add defaults to the struct definition rather than adding a constructor (I.e. `int id = -1`).

onboard/cv/main.cpp
```diff
@@ -108,30 +112,46 @@ int main() {
...
-        Point2f center = tennisBall.first;
-        float dist = depth_img.at<float>(center.y, center.x);
-        if (!isnormal(dist) || dist > 5.0) {
-          dist = (TENNIS_BALL_PIXEL_SIZE / tennisBall.second);
+      tagPair = detector.findARTags(src, depth_img);
+
+      //update both tags in LCM message
+      //first tag
+      if(tagPair.first.id == -1){//no tag found
+        if(left_tag_buffer <= 5){//send the buffered tag
```
amverni: I would do some testing in practice to see if the number 5 is suitable or not given what Caleb said about how often this code gets called (I would intuitively think this number needs to be higher).

onboard/cv/perception.hpp
```diff
@@ -1,6 +1,7 @@
 #pragma once
 #include <iostream>
 #include <opencv2/opencv.hpp>
+#include "opencv2/aruco.hpp"
```
amverni: why is this in quotes while `opencv2/opencv.hpp` is in angle brackets?

PR #290, opened 2020-02-08
[Nav] Radio Repeater Logic

elgarber:
Added logic to auton state machine to handle radio repeater drops

Added "Radio Repeater Drop" section to the Auton Task gui page (bool T or F whether the repeater has been dropped)

### DIFF
--- a/onboard/nav/rover.cpp
+++ b/onboard/nav/rover.cpp
@@ -282,7 +282,10 @@ void Rover::updateRepeater(RadioSignalStrength& radioSignal)
     static bool started = false;
     static time_t startTime;
 
-    if( !started &&
+    // If we haven't already dropped a repeater, the time hasn't already started
+    // and our signal is below the threshold, start the timer
+    if( !mTimeToDropRepeater &&
+        !started &&
         radioSignal.signal_strength <=
         mRoverConfig[ "radioRepeaterThresholds" ][ "signalStrengthCutOff" ].GetDouble())
     {
@@ -291,7 +294,7 @@ void Rover::updateRepeater(RadioSignalStrength& radioSignal)
     }
 
     double waitTime = mRoverConfig[ "radioRepeaterThresholds" ][ "lowSignalWaitTime" ].GetDouble();
-    if( difftime( time( nullptr ), startTime ) > waitTime )
+    if( started && difftime( time( nullptr ), startTime ) > waitTime )
     {
         started = false;
         mTimeToDropRepeater = true;
@@ -299,8 +302,6 @@ void Rover::updateRepeater(RadioSignalStrength& radioSignal)
 }
 
 // Returns whether or not enough time has passed to drop a radio repeater.
-// If it has been longer than the lowSignalWaitTime since a good signal strength,
-// return true. Otherwise, it is not time yet and we should keep waiting and return false.
 bool Rover::isTimeToDropRepeater()
 {
     return mTimeToDropRepeater;
--- a/onboard/nav/rover.cpp
+++ b/onboard/nav/rover.cpp
@@ -32,7 +32,7 @@ Course& Rover::RoverStatus::course()
 } // course()
 
 // Gets a reference to the rover's path.
-queue<Waypoint>& Rover::RoverStatus::path()
+deque<Waypoint>& Rover::RoverStatus::path()
 {
     return mPath;
 } // path()
@@ -59,6 +59,10 @@ Target& Rover::RoverStatus::target2() {
     return mTarget2;
 }
 
+RadioSignalStrength& Rover::RoverStatus::radio() {
+    return mSignal;
+}
+
 unsigned Rover::RoverStatus::getPathTargets()
 {
   return mPathTargets;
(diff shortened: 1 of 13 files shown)


onboard/nav/stateMachine.cpp
```diff
@@ -405,6 +493,25 @@ bool StateMachine::isWaypointReachable( double distance )
...
 
+bool StateMachine::isRadioRepeaterDrop() const
+{
+    if ( mPhoebe->roverStatus().currentState() != NavState::RadioRepeaterTurn &&
+         mPhoebe->roverStatus().currentState() != NavState::RadioRepeaterDrive &&
+         mPhoebe->isDropRepeater() &&
+         mRepeaterDropComplete == false &&
+         mRadioRepeaterDrops == 0 )
+    {
+        Waypoint way = (mPhoebe->roverStatus().course().waypoints)[mCompletedWaypoints-1];
```
amverni: I think you want to set `search` (and probably `gate`) to `false` here as well otherwise your logic will perform this if I'm following correctly.

onboard/nav/stateMachine.cpp
```diff
@@ -405,6 +493,25 @@ bool StateMachine::isWaypointReachable( double distance )
...
+        mPhoebe->roverStatus().path().push_front(way);
+        return true;
+    }
+    return false;
+}
+// void StateMachine::beginRepeaterDrop()
```
amverni: remove

onboard/nav/stateMachine.hpp
```diff
@@ -7,6 +7,7 @@
 #include "search/searchStateMachine.hpp"
 #include "obstacle_avoidance/simpleAvoidance.hpp"
 
+
```
amverni: remove

onboard/nav/stateMachine.hpp
```diff
@@ -99,6 +110,12 @@ class StateMachine
     // Number of waypoints completed.
     unsigned mCompletedWaypoints;
 
+    // Number of radio repeaters dropped.
+    unsigned mRadioRepeaterDrops;
```
amverni: It doesn't seem like this variable does anything. Multiple other parts of our code base would need to change if we added multiple drops so I think we should remove this.

onboard/nav/stateMachine.cpp
```diff
@@ -405,6 +493,25 @@ bool StateMachine::isWaypointReachable( double distance )
     return isLocationReachable( mPhoebe, mRoverConfig, distance, mRoverConfig["navThresholds"]["waypointDistance"].GetDouble());
 } // isWaypointReachable
 
+bool StateMachine::isRadioRepeaterDrop() const
```
amverni: comment function. I'm also confused by this naming. Is it checking something or performing an action?

onboard/nav/stateMachine.cpp
```diff
@@ -405,6 +493,25 @@ bool StateMachine::isWaypointReachable( double distance )
...
+{
+    if ( mPhoebe->roverStatus().currentState() != NavState::RadioRepeaterTurn &&
```
amverni: This is really complicated for an if statement so please comment what you're checking

onboard/nav/stateMachine.cpp
```diff
@@ -124,18 +138,26 @@ void StateMachine::run()
...
+            case NavState::RadioRepeaterDrive:
             case NavState::Drive:
             {
                 nextState = executeDrive();
                 break;
             }
 
+            case NavState::RadioRepeaterDrop:
+            {
+                nextState = executeRepeaterWait();
```
amverni: Please make the function and the state the same for consistency.

onboard/nav/rover.cpp
```diff
@@ -265,6 +273,15 @@ const double Rover::longMeterInMinutes() const
...
 }
 
+void Rover::updateRepeater(RadioSignalStrength& radioSignal)
+{
+    (radioSignal.signal_strength <= mRoverConfig[ "radioRepeater" ][ "lowSignal" ].GetDouble()) ? missedPings += 1 : missedPings = 0;
+}
+
+bool Rover::isDropRepeater()
+{
+    return (missedPings >= mRoverConfig[ "radioRepeater" ][ "threshold" ].GetDouble());
```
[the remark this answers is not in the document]
elgarber: To clarify. If we get a low signal, are you picturing that the rover stops moving for a certain amount of time to see if the signal is consistently low? or is it an internal clock and the rover can continue moving? Part of the reason I chose pings was because the rover could keep moving, but maybe that's not a good idea.

onboard/nav/stateMachine.cpp
elgarber: what do you mean by "set search and gate to false"? It can't go into either search state while in a Radio Repeater state.

elgarber: now I understand. Thank you!

onboard/nav/stateMachine.cpp
```diff
@@ -309,12 +332,26 @@ NavState StateMachine::executeTurn()
...
     if( mPhoebe->turn( nextPoint ) )
     {
+        if (mPhoebe->roverStatus().currentState() == NavState::RadioRepeaterTurn)
+        {
+            return NavState::RadioRepeaterDrive;
+        }
         return NavState::Drive;
     }
+
+    if (mPhoebe->roverStatus().currentState() == NavState::RadioRepeaterTurn)
```
amverni: Curly braces

onboard/nav/stateMachine.cpp
```diff
@@ -341,18 +387,47 @@ NavState StateMachine::executeDrive()
...
     }
     if( driveStatus == DriveStatus::OnCourse )
     {
+        if (mPhoebe->roverStatus().currentState() == NavState::RadioRepeaterDrive)
+        {
+            return NavState::RadioRepeaterDrive;
+        }
         return NavState::Drive;
     }
     // if driveStatus == DriveStatus::OffCourse
```
amverni: NIT: This comment is now confusing with the new logic below, could we make the comment say else so it's less misleading. Or remove it altogether. As is, it looks like it goes with the if statement and that's not true.

PR #291, opened 2020-02-09
[nav-sim] Gate/AR tag bug fixes and improvements

amverni:


simulators/nav/src/components/DrawModeModule.html
```diff
@@ -41,9 +41,9 @@
         <button disabled="{{gate_enabled}}" on:click="flip_target_ids()">
             Flip Target IDs
         </button>
-        <div class="{{gate_enabled ? 'disabled' : '' }}">
+        <div class="{{gate_enabled && wp_enabled ? 'disabled' : '' }}">
```
elgarber: Why is this enabled for both gate and waypoint? Was this for usability? The only time someone is going to change the gate width would be when setting down a gate, right? I guess it just seems weird for me that it's the only thing for gates that's enabled for waypoints too

amverni: Good catch. I will add a help tip to explain this so it's clear for users.

PR #295, opened 2020-02-12
[teleop] updated ra/sa arm controls

acfang:


base_station/gui/src/components/PidTune.vue
```diff
@@ -63,7 +65,7 @@
         Kp: 0.0,
         Ki: 0.0,
         Kd: 0.0,
-        Kf: 0.0,
+        pidMode: false, // true = sa, false = ra
```
jubeemer: Making this a boolean makes it sound like PID control is on or off, rather than switching between two configurations. Is there a better way to name this (i.e., `pidConfig`?) and make it something like an int value to represent state rather than a boolean value? (I don't know JS so if you disagree or I'm misunderstanding we can talk about it)

base_station/gui/src/components/SAControls.vue
```diff
@@ -119,6 +143,8 @@ export default {
     const esc_0_on = new Toggle(false)
     const esc_1_on = new Toggle(false)
 
+    document.addEventListener('keyup', this.keymonitor)
+
```
jubeemer: This looks like duplicate code from `ArmControls.vue`. Again, not familiar with JS/GUI design, but could a separate component be created for the Gimbal controls so we don't have to duplicate this?

onboard/teleop/src/__main__.py
```diff
@@ -38,6 +39,11 @@ def new_reading(self, reading):
 connection = None
 # front_drone_on = Toggle(False)
 # back_drone_on = Toggle(False)
+electromagnet_toggle = Toggle(False)
+laser_toggle = Toggle(False)
+solenoid_on = Toggle(False)
+electromagnet_on = Toggle(False)
+laser_on = Toggle(False)
```
jubeemer: These are left over from '18-'19, we won't need these anymore. If you want to keep them around for an example in case we end up with toggle controlled actuators, you could comment them out.

onboard/teleop/src/__main__.py
```diff
@@ -145,26 +143,58 @@ def drive_control_callback(channel, msg):
...
-        openLoopMsg.joint_id = i
-        openLoopMsg.speed = motorSpeeds[i]
-        lcm_.publish('/arm_motors', openLoopMsg.encode())
+    motor_speeds = [-deadzone(quadratic(xboxData.left_js_x), 0.09)*0.5,
+                    -deadzone(quadratic(xboxData.left_js_y), 0.09)*0.5,
+                    quadratic(xboxData.left_trigger -
+                              xboxData.right_trigger)*0.6,
+                    deadzone(quadratic(xboxData.right_js_y), 0.09)*0.75,
+                    deadzone(quadratic(xboxData.right_js_x), 0.09)*0.75,
+                    (xboxData.d_pad_right-xboxData.d_pad_left)*0.6]
```
jubeemer: The scalars are left over from a few years ago. They should have been removed last year but weren't. Please remove the scalars so that these will be capable of publishing in the full range [-1, 1].  Leave the `0.09` deadzone (I believe that's what that value is), it's probably useful to eliminate noise from bumping the joysticks.

onboard/teleop/src/__main__.py
```diff
@@ -222,13 +262,51 @@ def autonomous_callback(channel, msg):
         await asyncio.sleep(1)
 
 
+def sa_control_callback(channel, msg):
+    xboxData = Xbox.decode(msg)
+
+    saMotorsData = [-deadzone(quadratic(xboxData.left_js_x), 0.09)*0.5,
+                    -deadzone(quadratic(xboxData.left_js_y), 0.09)*0.5,
+                    quadratic(xboxData.left_trigger -
+                              xboxData.right_trigger)*0.6]
```
jubeemer: As explained above, please remove the scalars from these values.

base_station/gui/src/components/ArmControls.vue
```diff
@@ -78,59 +104,61 @@ export default {
...
+            //   this.$parent.publish('/arm_motors', openloop)
+
+            //   openloop.joint_id = 6
+            //   openloop.speed = (xboxData['right_bumper'] - xboxData['left_bumper'])
+            //   this.$parent.publish('/arm_motors', openloop)
+            // }
+
+            // if(send_arm_toggles) {
+            //   this.$parent.publish('/arm_toggles', arm_toggles)
+            // }
```
jubeemer: In general, remove old code rather than comment out big chunks like this. With version control, if we want it back at some point we can get it back. Leaving huge commented sections make s the source files long and harder to read through easily.

base_station/gui/src/components/SAControls.vue
```diff
@@ -159,22 +185,24 @@ export default {
...
+            //   'type': 'SAMotors',
+            //   'carriage': deadzone(xboxData['right_js_y'], 0.3),
+            //   'four_bar': xboxData['d_pad_left'] - xboxData['d_pad_right'],
+            //   'front_drill': front_drill_input,
+            //   'back_drill': back_drill_input,
+            //   'micro_x': deadzone(xboxData['left_js_x'], 0.3),
+            //   'micro_y': deadzone(xboxData['left_js_y'], 0.3),
+            //   'micro_z': xboxData['right_trigger'] - xboxData['left_trigger']
+            // }
+            // this.$parent.publish('/sa_motors', saMotorsData)
```
jubeemer: See above comment re: commenting out old code.

onboard/teleop/src/__main__.py
```diff
@@ -38,6 +39,11 @@ def new_reading(self, reading):
 connection = None
 # front_drone_on = Toggle(False)
 # back_drone_on = Toggle(False)
```
jubeemer: See above comment re: commenting out old code.

onboard/teleop/src/__main__.py
```diff
@@ -145,26 +143,58 @@ def drive_control_callback(channel, msg):
...
+    #     openloop = {
+    #         'type': 'OpenLoopRAMotor',
+    #         'joint_id': 5,
+    #         'speed': (xboxData.d_pad_right - xboxData.d_pad_left)*0.60,
+    #     }
+    #     lcm_.publish('/arm_motors', openloop)
+
+    #     openloop.joint_id = 6
+    #     openloop.speed = (xboxData.right_bumper - xboxData.left_bumper)
+    #     lcm_.publish('/arm_motors', openloop)
```
jubeemer: See above comment re: commenting out old code.

onboard/teleop/src/__main__.py
```diff
@@ -175,6 +205,16 @@ def arm_toggles_button_callback(channel, msg):
...
+    # arm_toggles = {
+    #     'type': 'ArmToggles',
+    #     'solenoid': xboxData.b,
+    #     'electromagnet': electromagnet_toggle.new_reading(xboxData.a),
+    #     'laser': laser_toggle.new_reading(xboxData.x)
+    # }
+
+    # if send_arm_toggles:
+    #     lcm_.publish('/arm_toggles', arm_toggles)
+
```
jubeemer: See above comment re: commenting out old code

onboard/teleop/src/__main__.py
```diff
@@ -222,13 +262,51 @@ def autonomous_callback(channel, msg):
...
+    openloop_msg.throttle = saMotorsData
+
+    lcm_.publish('/sa_openloop_cmd', openloop_msg.encode())
+
+    # lcm_.publish('/sa_openloop_cmd', saOpenLoopData)
+    # Can only toggle either front or back drill
+    # let front_drill_input = this.controlMode ==
+    #   'front_drill' and xboxData.right_bumper
+    # let back_drill_input = this.controlMode ==
+    #   'back_drill' and xboxData.right_bumper
```
jubeemer: See above comment re: commenting out old code

PR #301, opened 2020-02-20
[nav] SAR critical components

amverni:
This includes initial radio repeater logic and initial gate traversal logic


amverni:
The TODOS relate to three things. First, we need to add the `GateShimmy` (name up in the air at the moment) nav state in order to center ourselves more accurately but the current version works just not as well as we would like it to in the end. Second, we ignored if there is an obstacle in very close proximity to the gate (namely in the middle of the gate). While this is something we would want to avoid, we think this is such an extreme edge case that ignoring it for now is not a costly mistake. Third, if we go "off course" when trying to go through a gate, we just keep driving towards our goal. Since driving through the gate is not SAR critical we did not spend the time perfecting this portion right now but we feel it must be taken care of in the future. This will also be mitigated when we implement `GateShimmy` and we can add similar logic once we perfect that algorithm.

PR #317, opened 2020-06-07
[nav] Refactor navigation simulator

### LINKED ISSUE #308: Refactor navigation simulator to use Vue and Typescript

amverni:
resolves #308

Refactor navigation simulator to use Vue and TypeScript.
This refactor is indented to make the simulator easier to maintain,
make it easier to add new features, and move towards using
technologies that are more commonly used in industry.

We are keeping around the old version of the simulator
temporarily in case the refactored version has any major
problems. This should be deleted when the correctness of the
refactored version is confirmed.


amverni:
There was a bug in the radio signal LCM message being published. I'm a bit concerned why this worked sometimes and not others. If you were getting a segfault from Nav previously, this was likely triggering it (however it was not the cause).

DavidSmith166:
It looks like you rebased or amended your history after fixing the last bug. This makes it difficult to review what the most recent changes you made were. In the future, leave the history cleanup until the end so as to provide the highest level of transparency to reviewers.

PR #333, opened 2020-08-28
[Localization] Extend GPS capability with Real-Time Kinematics 

DavidSmith166:
Includes rewrite of onboard/gps program, introduction of base_station/gps program, modification of GPS lcm struct, and addition of RTCM lcm struct.

[What is RTK?](https://novatel.com/an-introduction-to-gnss/chapter-5-resolving-errors/real-time-kinematic-rtk)


base_station/gui/src/components/LCMEcho.vue
```diff
@@ -82,6 +82,7 @@
           '/ra_pidconfig_cmd': false,
           '/rr_drop_complete': false,
           '/rr_drop_init': false,
+          '/rtcm': false,
```
amverni, line 85: What is rtcm? Can we make this something more readable? I know other channels probably don't have the most readable names either but no need to continue bad habits.

Same goes for the LCM struct and adding at least a comment when rtcm gets introduced in the python program.

base_station/gps/src/__main__.py
```diff
@@ -0,0 +1,51 @@
+import serial
+import time
+import json
+from os import getenv
+from rover_common import aiolcm
+from rover_msgs import RTCM
+
+
+class GPS_Base_Station():
+
```
amverni, line 10: NIT: Remove?

rover_msgs/GPS.lcm
```diff
@@ -6,5 +6,14 @@ struct GPS {
     int32_t longitude_deg;
     double longitude_min;
     double bearing_deg;
+    // Units: degrees from true
+    // 999: Invalid/Nonexistent measurement
```
amverni, line 10: If we are going to make individual fields take up multiple lines (including comments) can you put a newline between them? It's not clear to me what comments go to what field in all cases.

onboard/gps/src/__main__.py
```diff
@@ -0,0 +1,162 @@
...
+from os import getenv
+from rover_common.aiohelper import run_coroutines
+from rover_common import aiolcm
+from rover_msgs import GPS, RTCM
+
+
+class GPS_Manager():
+
+    def __init__(self):
+
```
amverni, line 14: NIT: remove? I noticed you're consistent with this. Is this needed for the linter? If this is considered good style, then go ahead and leave it.

onboard/gps/src/__main__.py
```diff
@@ -0,0 +1,162 @@
+import serial
```
amverni, line 1: Can you add a docstring to the top of the file and for each function explaining what is going on from a high level or what is the purpose of the file/class/function?

base_station/gps/src/__main__.py
```diff
@@ -0,0 +1,51 @@
+import serial
```
amverni, line 1: Can you add a docstring to the top of the file and for each function explaining what is going on from a high level or what is the purpose of the file/class/function?

base_station/gui/src/components/LCMEcho.vue
DavidSmith166: [RTCM](https://www.rtcm.org/) stands for the Radio Technical Commission for Maritime Services of which have standardized the messages that a GPS use to coordinate with other devices. We use the RTCM messages that our base station GPS provides to acquire our RTK fix.

Similar to how we don't spell out the acronyms for "GPS", "IMU", or "LCM" I don't see why it would be valuable to name the channel for RTCM messages something other than RTCM. 

If you feel that it is necessary I could append "_msg" to the path, would that be sufficient?

base_station/gps/src/__main__.py
DavidSmith166: Could you be more clear as to what you want me to remove? Is it the white space or the class definition? The white space is following PEP8 format due to our linter, and I don't see an issue with using a class to provide additional functionality.

base_station/gps/src/__main__.py
DavidSmith166: lmk if I have addressed this sufficiently in my latest commit

rover_msgs/GPS.lcm
DavidSmith166: good call

base_station/gui/src/components/LCMEcho.vue
amverni: Ah okay. I still think a comment indicating this is for the base station GPS would be nice. Especially in the base station code where it is less obvious.

base_station/gps/src/__main__.py
amverni: I was talking about the whitespace. I attempted to comment on just the one line but it doesn't seem like that was made clear.

onboard/gps/src/__main__.py
amverni: Again, I was talking about the whitespace which we discussed earlier is for the linter.

base_station/gps/src/__main__.py
amverni: I would still push for commenting the classes and functions but I will also accept it as it is now.

DavidSmith166: What level of granularity are you looking for in the documentation? I am open to writing more especially if any part of my code is unclear

onboard/gps/src/__main__.py
```diff
@@ -121,12 +127,14 @@ async def recieve(self, lcm):
                 continue
 
             lcm.publish('/gps', gps_struct.encode())
-            seen_tags = {tag: False for tag in self.NMEA_TAGS_MAPPER.keys()}
+            seen_tags = {tag: False if not tag == 'TXT' else True
+                         for tag in self.NMEA_TAGS_MAPPER.keys()}
```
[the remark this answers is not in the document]
DavidSmith166: Actually the opposite, the only tag that I am not looking for is TXT tags, but I have a handler for them in case they show up and the way I have things setup is such that every supported tag is expected to show up regularly so I made a quick hack to exclude TXT. And for clarity TXT statements hold error msgs and that sort of thing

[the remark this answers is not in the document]
raytitan: This makes sense

raytitan (approved):

rzf16 (approved):
🥇

PR #382, opened 2020-09-21
Imu

cgiger00:
IMU code for mock comp - does not use DMP run time calibration, using the arduino tool method


cgiger00:
onboard filter is failing because it uses IMU.lcm not IMUData.lcm

PR #394, opened 2020-09-26
[lcm] Added an lcm struct for messages intended for operators

raytitan:
This same structs should be used by any rover program on the "/debug" "/warning" and "/critical" channels to notify the operator of rover events


amverni:
Would it be more useful to have `debug`, `warning`, or `critical` be in the LCM message rather than the channel name? For example in a `level` variable? I think this would be more useful in the case that someone was looking at the echo page for distinguishing messages. Then you could put them on the same channel allowing someone to see all messages from the command line lcm echo rather than needing multiple terminals. If the problem is wanting to only see one at a time, would a filter on the echo page be worthwhile?

amverni:
Do we want to move this PR to a different branch (such as a system admin branch) in order to have a consistent workflow rather than having this be an exception that PRs directly to `main`?

raytitan:
> 
> 
> Would it be more useful to have `debug`, `warning`, or `critical` be in the LCM message rather than the channel name? For example in a `level` variable? I think this would be more useful in the case that someone was looking at the echo page for distinguishing messages. Then you could put them on the same channel allowing someone to see all messages from the command line lcm echo rather than needing multiple terminals. If the problem is wanting to only see one at a time, would a filter on the echo page be worthwhile?

I think I agree with you on this

rover_msgs/RoverLog.lcm
```diff
@@ -0,0 +1,12 @@
+package rover_msgs;
+
+struct RoverStatus {
+
+    uint8_t CRITICAL = 0; //Requires operator action. Provide basic recovery steps in <message>
+    uint8_t WARNING = 1; //Requires operator attention. Program should attempt self recovery
+    uint8_t DEBUG = 2; //Basic, useful information. No action necessary
+
+    uint8_t priority; //Set this to RoverStatus.CRITICAL, .WARNING, or .DEBUG
```
amverni, line 9: Why did you choose to use a `uint8_t` (i.e. an enum) instead of a `string`? The nav status LCM used to do the same but we ended up having to store the enum definition in at least 3 locations. This meant when the possible values changed, you had to make sure to change all of them which is arguably not good style. On a less important note, this also requires users to memorize the mapping from number to meaning as the LCM echo as is won't translate it.

amverni:
How is this different than the `DebugMessage` LCM that already exists?

rover_msgs/RoverLog.lcm
raytitan: With lcm structs you don't have to store the enum, the enum is stored in the lcm message's namespace. So you can say RoverLog.priority = RoverLog.CRITICAL.

raytitan:
> 
> 
> How is this different than the `DebugMessage` LCM that already exists?
I was unaware that this message exists, seems to mostly used by teleop. Perhaps DebugMessage could be repurposed? (It doesn't have the originator and the priority fields, which I think are better to enforce through the data structure than through usage guidelines)

rover_msgs/RoverLog.lcm
amverni: This method can't be done for JavaScript services (e.g. the base station), can it?

PR #401, opened 2020-09-27
[esw] thermistor

cgiger00:
as tested as it can be without actually running inside the science boxes


beaglebone/thermistor/src/__main__.py
```diff
@@ -0,0 +1,167 @@
...
+    lcm = lcm_.LCM()
+
+    # init the ADC
+    adc.setup()
+
+    lcm.subscribe("/thermistor_cmd", request_callback)
+
+    # Setting up lock for outputting
+    global outLock
+    outLock = threading.Lock()
```
raytitan, line 156: What's the point of this lock?

outlock.lock() would force other threads that call outlock.lock() to wait, but I don't see any other functions using it

raytitan (commented):
We should have a discussion later about the pros/cons of just publishing data from all sensors without listening for cmd lcms

PR #406, opened 2020-09-28
[esw] servo code update

cgiger00:
servos 0 & 1 were switched in wiring the science box - updated the code to reflect that


cgiger00:
merging!

PR #407, opened 2020-09-28
Perception added initial point cloud implementation

mmartin4972:
I have a lot of #if statements to isolate the obstacle detection stuff and build without it, so it passes Travis. I have people working on updating the Dockerfile, so that we will be able to actually test to see if more than 40% of our code compiles in Travis. They can't test that however, unless they have my PCL changes.


raytitan (commented):
Have we always stored CV test images in the mrover-workspace repo?

Also would love to see jetson/percep using command line arguments instead of compilation flags to distinguish between situations like debug mode, ARTag on/off, PCL Obstacle on/off, etc...., to be more flexible

mmartin4972:
> Have we always stored CV test images in the mrover-workspace repo?
> 
> Also would love to see jetson/percep using command line arguments instead of compilation flags to distinguish between situations like debug mode, ARTag on/off, PCL Obstacle on/off, etc...., to be more flexible

We have always stored cv_test images in the workspace. They were under a different folder that was rarely used. Now with everyone doing things remote they are going to be reading in test images much more, so I figured it's kind of nice for them to just be in the workspace.

We can definitely start moving towards command line arguments. Some things like whether or not we are running with the ZED would need compilation flags because of the libraries it includes during build. Point Cloud obstacle is the same way with specifying libraries, so we'll probably want to keep that a compilation flag. Other than that we should be able to migrate over to command line args in the future.

DavidSmith166 (approved):

Issue #355, opened 2020-09-14
Deprecate the detected variable in the Obstacle message.


amverni:
- [x] Remove `detected` from LCM `Obstacle` message
- [x] Remove `detected` from `Obstacle` message published by `simulators/nav`
- [x] Remove `detected` from `Obstacle` message published by `onboard/cv`

`simulators/nav` already does not rely on this variable (e.g. it is always set to false). This change is dependent `onboard/nav` and `onboard/cv` also not using this variable.

PR #417, opened 2020-10-01
fix sim bug and update interface for obstacle detection

aswarner:
Merge deprecatedetected from aswarner into nav-main. This commit resolves nav sim interface issues by updating nav and perception code for obstacle lcm.  The lcm conventions for AR tag and obstacle are standardized.
-detected boolean was removed from lcm
-nav code was updated to interpret obstacle distance = -1 as no obstacle detected
-cv code was updated to send distance = -1 when no obstacle detected


onboard/cv/main.cpp
```diff
@@ -163,13 +163,12 @@ int main() {
...
       if(obstacle_detection.bearing > 0.05 || obstacle_detection.bearing < -0.05) {
         // cout<< "bearing not zero!\n";
-        obstacleMessage.detected = true;    //if an obstacle is detected in front
         obstacleMessage.distance = obstacle_detection.center_distance; //update LCM distance field
       }
       obstacleMessage.bearing = obstacle_detection.bearing;
 
       #if PERCEPTION_DEBUG
-      // cout << "Turn " << obstacleMessage.bearing << ", detected " << (bool)obstacleMessage.detected<< endl;
+      // cout << "Turn " << obstacleMessage.bearing << ", distance " << (bool)obstacleMessage.distance << endl;
```
amverni, line 171: Why is the `#if` here and the code is commented out?

onboard/nav/obstacle_avoidance/simpleAvoidance.cpp
```diff
@@ -102,3 +102,9 @@ Odometry SimpleAvoidance::createAvoidancePoint( Rover* phoebe, const double dist
 
 
 } // createAvoidancePoint()
+
+// Returns true if an obstacle is detected, false otherwise.
+bool SimpleAvoidance::isObstacleDetected(Rover * phoebe) const
```
amverni, line 107: Don't we already have an `isObstacleDetected()` function?

onboard/nav/stateMachine.cpp
```diff
@@ -489,7 +489,7 @@ string StateMachine::stringifyNavState() const
 // Returns true if an obstacle is detected, false otherwise.
 bool StateMachine::isObstacleDetected() const
```
amverni, line 490: Why is this function duplicated with the one in `searchStateMachine`?

amverni (changes requested):
Couple high level comments:

1. Make sure you squash the commits after review based on our new workflow process.
2. These code changes aren't just nav so the final commit message shouldn't imply that.
3. Why is there a merge commit in here? I don't think you need that.
4. How was this tested?

amverni:
Your comment above is good! You should include that in the final commit message rather than just a comment on this PR.

onboard/cv/main.cpp
mmartin4972: Our code needs some cleansing

mmartin4972:
I'm pretty sure you guys changed everything that needed it. From a Perception standpoint this seems fine

onboard/nav/stateMachine.cpp
aswarner: because the isObstacleDetected() function is a private member function of stateMachine, which is not inherited by gate state machine or searchStateMachine. If you think it's best, I can make it just one public function. it didn't occur to me to change this because the powers that be had implemented it as a private function

onboard/cv/main.cpp
aswarner: I pulled from percep-main so now the file should include the recent changes from Matt as well as my fix

mmartin4972:
@aswarner can you add "resolves #355" in the commit message?

mmartin4972 (changes requested):
From what I can tell it seems like you guys changed everything. If not we can always adjust it on Perception's end. Could you add "resolves #355" to one of the commit messages?

jetson/nav/stateMachine.cpp
```diff
@@ -489,7 +489,7 @@ string StateMachine::stringifyNavState() const
 // Returns true if an obstacle is detected, false otherwise.
 bool StateMachine::isObstacleDetected() const
 {
-    return mPhoebe->roverStatus().obstacle().detected;
+    return mPhoebe->roverStatus().obstacle().distance >= 0;
```
atguimond, line 492: I am in favor of abstracting this out into a function, as long as you have it that way in simpleAvoidance, just to keep it consistent.

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -281,7 +281,7 @@ NavState SearchStateMachine::executeDriveToTarget( Rover* phoebe, const rapidjso
 // Returns true if an obstacle is detected, false otherwise.
 bool SearchStateMachine::isObstacleDetected( Rover* phoebe ) const
 {
-    return phoebe->roverStatus().obstacle().detected;
+    return phoebe->roverStatus().obstacle().distance >= 0;
```
atguimond, line 284: Also in favor of abstracting this out.

atguimond (approved):

Issue #362, opened 2020-09-15
Add hints for hotkeys


amverni:
Option 1: Add popups when hovering over buttons to say what their corresponding hotkeys are.
Option 2: Have a help menu

- [ ] Reset rover - `shift+r`
- [ ] Pause/play - `shift+space`
- [ ] Take step - `shift+alt+space`
- [ ] On/Off - `shift+enter`
- [ ] Switching draw modes - `shift+[1-5]`
- [ ] Clear field - `shift+backspace`
- [ ] Simulate localization - `shift+l`
- [ ] Simulate perception - `shift+p`
- [ ] Toggle gate search - `shift+g`
- [ ] Toggle search - `shift+s`
- [ ] Switch odom format - `shift+alt+[dms]`
- [ ] #395 

See #431 for complete list of current hotkeys

PR #422, opened 2020-10-01
[nav-sim] Add ability to move rover and start point

### LINKED ISSUE #395: Adjust rover starting point
- [x] While holding shift+c and clicking on the field, instead of adding a field item, change the rover's current location. 
- [x] While holding shift+c and double click on the field to change the rover's starting location and current location.
- [x] Clicking shift+ctrl+c resets the starting location and moves the rover to that location.
- [x] Clicking shift+alt+c to set the starting odom to the rover's current odom.

amverni:
resolves #395

Add hotkeys to move the rover by clicking on the simulator
field, adjust the rover's starting position, and reset the
rover's starting position.

### DIFF
--- a/simulators/nav/src/components/HotKeys.vue
+++ b/simulators/nav/src/components/HotKeys.vue
@@ -12,7 +12,12 @@
 <script lang="ts">
 import { Component, Vue } from 'vue-property-decorator';
 import { Getter, Mutation } from 'vuex-class';
-import { FieldItemType, Joystick, OdomFormat } from '../utils/types';
+import {
+  FieldItemType,
+  Joystick,
+  Odom,
+  OdomFormat
+} from '../utils/types';
 
 @Component({})
 export default class HotKeys extends Vue {
@@ -23,7 +28,13 @@ export default class HotKeys extends Vue {
   private readonly autonOn!:boolean;
 
   @Getter
+  private readonly currOdom!:Odom;
+
+  @Getter
   private readonly drawMode!:FieldItemType;
+
+  @Getter
+  private readonly fieldCenterOdom!:Odom;
 
   @Getter
   private readonly paused!:boolean;
@@ -38,7 +49,13 @@ export default class HotKeys extends Vue {
    * Vuex Mutations
    ************************************************************************************************/
   @Mutation
+  private readonly clearRoverPath!:()=>void;
+
+  @Mutation
   private readonly setAutonState!:(onOff:boolean)=>void;
+
+  @Mutation
+  private readonly setCurrOdom!:(newOdom:Odom)=>void;
 
   @Mutation
   private readonly setDrawMode!:(mode:FieldItemType)=>void;
@@ -53,6 +70,9 @@ export default class HotKeys extends Vue {
   private readonly setPaused!:(paused:boolean)=>void;
 
   @Mutation
+  private readonly setStartLoc!:(newStartLoc:Odom)=>void;
+
+  @Mutation
   private readonly setTakeStep!:(takeStep:boolean)=>void;
 
   @Mutation
@@ -62,10 +82,16 @@ export default class HotKeys extends Vue {
   private readonly flipSimulatePercep!:(onOff:boolean)=>void;
 
   /************************************************************************************************
+   * Private Members
+   ************************************************************************************************/
+  /* Whether or not clicking on the field recenters the field. */
+  private prevDrawMode:FieldItemType|null = null;
+
+  /************************************************************************************************
    * Local Getters/Setters
    ************************************************************************************************/
   /* Mapping of hotkeys to functions. */
-  get keymap():Record<string, ()=>void> {
+  get keymap():Record<string, (()=>void)|(Record<string, ()=>void>)> {
     return {
       'shift+enter': this.flipAutonState,
       'shift+space': this.pausePlay,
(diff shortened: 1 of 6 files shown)


aswarner:
which keys are the hotkeys? I'm unsure of how to test this

amverni:
> which keys are the hotkeys? I'm unsure of how to test this

#395 and #362

simulators/nav/src/components/HotKeys.vue
```diff
@@ -61,11 +81,17 @@ export default class HotKeys extends Vue {
...
+   ************************************************************************************************/
+  /* Whether or not clicking on the field recenters the field. */
+  private prevDrawMode:FieldItemType|null = null;
+
   /************************************************************************************************
    * Local Getters/Setters
    ************************************************************************************************/
   /* Mapping of hotkeys to functions. */
-  get keymap():Record<string, ()=>void> {
+  get keymap():Record<string, (()=>void)|(Record<string, ()=>void>)> {
```
DavidSmith166, line 94: I think a wiki page detailing how the sim works would be a great place to enumerate these hotkeys

simulators/nav/src/components/HotKeys.vue
```diff
@@ -143,6 +175,15 @@ export default class HotKeys extends Vue {
     this.setPaused(!this.paused);
   } /* pausePlay() */
 
+  /* Move the start odom back to the field center. */
```
DavidSmith166, line 178: NIT: perhaps this could be more clear, is the odom the position of the rover?

simulators/nav/src/components/field/Field.vue
```diff
@@ -347,22 +370,51 @@ export default class Field extends Vue {
...
-    this.dataToDownload.removeAttribute('href');
-    this.dataToDownload.removeAttribute('download');
-  } /* downloadTestCase() */
+    switch (this.drawMode) {
+      case FieldItemType.MOVE_ROVER: {
+        this.moveStartOdom(clickOdom);
+        break;
+      }
+
+      /* no default */
```
DavidSmith166, line 398: Is there no default because the default is to do nothing or is a drawMode that reaches the default case undefined behavior?

amverni: The default is to do nothing. The reason for this comment is readability: https://eslint.org/docs/rules/default-case

simulators/nav/src/components/HotKeys.vue
amverni: I think that's really valid. I think that would be an independent change. What do you think?

amverni: #431

aswarner:
Not related to this PR: why do we have an option to not simulate localization? It seems like the rover doesn't move at all without it. What use case does that represent?

aswarner:
For hotkey shift-alt-space: on [Windows?] browsers, alt-space opens a menu for opening/closing tabs, restoring, etc. The shift-alt-space hotkey did not work because the alt-space in the browser overrode our custom command. Can that hotkey be modified  to something without a conflict?

aswarner (commented):
The #395 hotkeys work, but I [belatedly] noticed a previously pushed bug from #362 :

For hotkey shift-alt-space: on [Windows?] browsers, alt-space opens a menu for opening/closing tabs, restoring, etc. The shift-alt-space hotkey did not work because the alt-space in the browser overrode our custom command. Can that hotkey be modified to something without a conflict? Not sure if that is appropriate for this PR or requires a different PR.

simulators/nav/src/components/HotKeys.vue
DavidSmith166: Personally I like the idea of shipping up-to-date  documentation with new features. I'd like to see #431 resolved with this PR as well

amverni: This goes against our 1 issue = 1 commit = 1 PR guideline which I would like to keep for a few reasons primarily based around making reviewing, implementing, and deploying changes faster and easier.

amverni:
> Not related to this PR: why do we have an option to not simulate localization? It seems like the rover doesn't move at all without it. What use case does that represent?

This will allow for testing with the real localization rather than simulated localization. In the future you will be able to input waypoints via GPS coordinates rather than just clicking on the screen and this will be more useful. However, in the meantime, you can turn off localization, run the localization service, move the start location to the current location, and then start testing from there.

amverni:
> The #395 hotkeys work, but I [belatedly] noticed a previously pushed bug from #362 :
> 
> For hotkey shift-alt-space: on [Windows?] browsers, alt-space opens a menu for opening/closing tabs, restoring, etc. The shift-alt-space hotkey did not work because the alt-space in the browser overrode our custom command. Can that hotkey be modified to something without a conflict? Not sure if that is appropriate for this PR or requires a different PR.

Create an issue for this and put this description in the body of the issue.

aswarner (dismissed):

simulators/nav/src/components/HotKeys.vue
DavidSmith166: Documentation is a necessary part of the development process and integral to the continued success of the team as our code base grows in size and complexity. As our documentation catches up to our code I will expect code changes to be paired with documentation changes in a single issue. This is to prevent our documentation from ever being out of date. Since there is nothing written instructing other people on rover how to use this code, I suppose there is no harm in it arriving at a later date.

amverni: I completely agree. I think I should have done a better job when I first made the hotkey feature about this documentation, but unfortunately I didn't think of it then. I will make this documentation my next priority.

DavidSmith166:
> In the future you will be able to input waypoints via GPS coordinates rather than just clicking on the screen and this will be more useful.

Has there been an issue created for this?

DavidSmith166 (approved):

amverni:
> > This will allow for testing with the real localization rather than simulated localization. In the future you will be able to input waypoints via GPS coordinates rather than just clicking on the screen and this will be more useful. However, in the meantime, you can turn off localization, run the localization service, move the start location to the current location, and then start testing from there.
> 
> Has there been an issue created for this?

#312

DavidSmith166 (approved):

PR #426, opened 2020-10-02
adding PCL comment to make pull request

nataliemckee:
Pull request to remain open as the Dockerfile is being edited to include PCL libraries.


amverni:
Can you make your commit message more clear and match the style outlined in the MRover workflow document? Also could you make the title of this pull request more descriptive as well. A change like this seems meaningless so having a good description is important.

nataliemckee:
This pull and commit were created to that I can edit the Dockerfile to install PCL libraries-- I will be adding a line, checking that it passes in the pull request check, and continuing to add lines until the PCL libraries may be installed successfully. Matt will not accept the pull request until that point, so there will be many more commits by then. I can rename the pull request.

amverni:
So my understanding is you have this PR just so you can test with Travis, correct? If that is the case, let's close this PR until it is closer to being done and add Travis to your fork. To do this, you just have to add a branch rule like `umrover/mrover-workspace` has. I have set this up for my fork as well and it's pretty easy to do.

PR #459, opened 2020-10-15
[auton] update sim functionality for integration branch

aswarner:
Previously the nav sim didn't work with auton integration code because of the deprecated obstacle lcm that is still used in auton integration code. This change will help nav test code that is currently on the branch to detect and resolve integration bugs.

When the nav-main code is eventually merged with the auton integration branch, the nav sim, nav state machine, and cv code will be updated to remove the deprecated obstacle lcm detected bool. (this update is already in nav-main)

### DIFF
--- a/simulators/nav/src/components/perception/obstacle_detector.ts
+++ b/simulators/nav/src/components/perception/obstacle_detector.ts
@@ -238,11 +238,20 @@ export default class ObstacleDetector {
       angle = intervalHeap.maxOccupied + (minIntervalSize / 2);
     }
 
-    return {
-      detected: false,
-      distance: this.obsDist,
-      bearing: angle
-    };
+    if (this.obsDist >= 0) {
+      return {
+        detected: true, /* deprecated so always false */
+        distance: this.obsDist, /* Will be -1 if okay to go straight ahead (i.e. bearing = 0) */
+        bearing: angle
+      };
+    }
+    else {
+      return {
+        detected: false, /* deprecated so always false */
+        distance: this.obsDist, /* Will be -1 if okay to go straight ahead (i.e. bearing = 0) */
+        bearing: angle
+      };
+    }
   } /* computeObsMsg() */
 
   /************************************************************************************************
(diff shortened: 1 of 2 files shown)


simulators/nav/src/components/perception/obstacle_detector.ts
```diff
@@ -238,11 +238,20 @@ export default class ObstacleDetector {
...
+        bearing: angle
+      };
+    }
+    else {
+      return {
+        detected: false, /* deprecated so always false */
+        distance: this.obsDist, /* Will be -1 if okay to go straight ahead (i.e. bearing = 0) */
+        bearing: angle
+      };
+    }
```
amverni, line 254: 1. Can you add a comment signifying that we do not want to add this to our repository for real? In other words this should not be merged into neither `main` nor any `*-main` branches.
2. Why do we want to do this? Isn't the right answer to correct the programs misusing this field rather than hack the program correctly using it?
3. what about:
```
return {
    detected: this.obsDist >= 0,
    distance: this.obsDist,
    bearing: angle
};
```

simulators/nav/src/components/perception/obstacle_detector.ts
```diff
@@ -279,12 +288,20 @@ export default class ObstacleDetector {
...
+        distance: this.obsDist, /* Will be -1 if okay to go straight ahead (i.e. bearing = 0) */
+        bearing: calcRelativeBearing(this.zedOdom.bearing_deg, angle)
+      };
+    }
+    else {
+      return {
+        detected: false, /* deprecated so always false */
+        distance: this.obsDist, /* Will be -1 if okay to go straight ahead (i.e. bearing = 0) */
+        bearing: calcRelativeBearing(this.zedOdom.bearing_deg, angle)
+      };
```
amverni, line 303: I don't think you need to do this twice. You are only affected by the version of these that is will eventually be published via the LCM, right? Could be this one or the other, but not both is my understanding.

DavidSmith166:
> 
> 
> Previously the nav sim didn't work with auton integration code because of the deprecated obstacle lcm that is still used in auton integration code. This change will help nav test code that is currently on the branch to detect and resolve integration bugs.
> 
> When the nav-main code is eventually merged with the auton integration branch, the nav sim, nav state machine, and cv code will be updated to remove the deprecated obstacle lcm detected bool. (this update is already in nav-main)

What is blocking nav-main from being merged into auton-integration?

amverni:
Closing due to inactivity. Additionally, a PR is not required for changes to this branch.

PR #461, opened 2020-10-25
negative longitude minutes bug fix

atguimond:


obstacle_lcm_output.txt
```diff
@@ -0,0 +1,1768 @@
+----- 2020-10-16 00:46:03.445463
+{'bearing_deg': 6.469894763364437,
+ 'latitude_deg': 38,
+ 'latitude_min': 24.383276087527292,
+ 'longitude_deg': -110,
+ 'longitude_min': -47.51196922390136,
+ 'speed': -1.0}
```
DavidSmith166, line 7: We shouldn't be saving txt files to our code repository, they can live in the google drive

DavidSmith166 (changes requested):

rzf16 (commented):
See this is why you never ask for reviews, that way David can't reject your changes

PR #494, opened 2020-11-19
[Percep] Improved ground-plane detection

niwhsa9:
Greatly improved ground-plane detection using the PCL perpendicular plane segmentation model instead of the plane model. Corrected the axis being set (+y is up in the camera frame). Further improvement can be achieved by tuning the epsilon angle as needed.

### DIFF
--- a/jetson/percep/pcl.cpp
+++ b/jetson/percep/pcl.cpp
@@ -51,13 +51,13 @@ void PCL::RANSACSegmentation(string type) {
     //Creates instance of RANSAC Algorithm
     pcl::SACSegmentation<pcl::PointXYZRGB> seg;
     seg.setOptimizeCoefficients(true);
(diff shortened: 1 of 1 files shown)


jetson/percep/pcl.cpp
```diff
@@ -459,7 +460,7 @@ obstacle_return PCL::pcl_obstacle_detection(shared_ptr<pcl::visualization::PCLVi
     obstacle_return result;
     PassThroughFilter();
     DownsampleVoxelFilter();
-    RANSACSegmentation("remove");
+    RANSACSegmentation("blue");
```
mmartin4972, line 463: Can you change this back to remove?

mmartin4972 (changes requested):

mmartin4972 (approved):

PR #502, opened 2020-11-20
[nav] Updating integration to match main

atguimond:
- Gets rid of deprecated obstacle boolean
- Updates simulator package
- Resolves Issue #498


amverni:
Something doesn't seem right; there are a lot of duplicated commits.

amverni:
I don't think you want to be squashing all of these commits into one thus this PR is not ideal.

atguimond:
I do not see duplicated commits. And why should I not be squashing these commits into one?

amverni:
Can you explain why you _do_ want to squash these commits?

atguimond:
I was just trying to streamline in all of these commits since they have already been made in other places. I can split them up into smaller and more logical commits if that would be better though if you think that would be better.

PR #527, opened 2020-12-06
[nav-sim] Remove old nav simulator

### LINKED ISSUE #487: Remove old simulator

amverni:
resolves #487 

Tested by verifying that the new simulator still works properly after deleting the old simulator.


raytitan (commented):
Is there a reason you haven't removed the rest of the files in this directory? Will travis (our automated tester) fail on this?

amverni:
> Is there a reason you haven't removed the rest of the files in this directory? Will travis (our automated tester) fail on this?

No. That's a really good catch! I forgot to remove those files.

raytitan (approved):
Given you've built and run the new sim fine, this looks good to me!

atguimond (approved):
Looks good! Is there any reason to keep the old sim around somewhere not in the repository in case the new sim breaks?

John-Trager:
Looks good to me!

amverni:
> Looks good! Is there any reason to keep the old sim around somewhere not in the repository in case the new sim breaks?

If we wanted to do this, we should just keep the old sim around. However, we don't want to do this because we are not maintaining the old sim so newer versions of the nav service won't work with the old sim anymore.

amverni:
It was accidentally closed.

John-Trager (approved):
Looks good!

Issue #528, opened 2020-12-06
[nav] Add ZED data lcm message


atguimond:

aswarner:
@ankithu @jchen3652 this issue will be part of your PR. when you make the PR, add "resolves #528" in your PR description to close this issue

Issue #471, opened 2020-11-01
Add ZED camera gimbal functionality to simulator


atguimond:
- Transferring rover search spin from spinning the wheels to spinning the camera gimbal
- Gimbal located on the front of the ebox
- Field of vision changes with gimbal
- Eliminating the need for spinning but not turning

amverni:
@atguimond Can you add a link to the PR and line number where the new/changed LCM message is allowing `jetson/nav` to control the gimbal? Also, please add the relevant labels and project for this issue.

atguimond:
This code is in testing stages so there is no PR yet. We were waiting until we could test it. The code is currently on this branch: https://github.com/ankithu/mrover-workspace/tree/gimbal-nav

amverni:
Got it. To complete the loop, [this](https://github.com/ankithu/mrover-workspace/blob/gimbal-nav/rover_msgs/GimbalCmd.lcm) is the interface we will be using to control the gimbal.

atguimond:
We created a different lcm message [here](https://github.com/ankithu/mrover-workspace/blob/gimbal-nav/rover_msgs/ZedGimbalCmd.lcm).

amverni:
Why are we creating a new LCM? Can we not reuse the old one? If not, what is the old one being used for and can we change it?

atguimond:
We created the new one to keep things simple and streamlined. I have no idea what the old one is for (probably another gimbal) and I did not want to step on anyone's toes if someone else is still using it. We also added a new lcm that ESW is publishing and nav is listening to called ZedGimbalData.lcm that sends the same stuff to us that we are sending back.

amverni:
Can you clarify what the power means? In other words. If I, acting as the simulator, get a power or `x`, what should I do with that? How would I interpret that? What range of values are valid? How fast does a power value of `x` move the gimbal (might not be able to answer that one)?

atguimond:
We refactored the code to make it easier in terms of yaw [here](https://github.com/ankithu/mrover-workspace/blob/gimbal-nav-no-pid/rover_msgs/ZedGimbalCmd.lcm). 0 degrees is straight ahead, +90 is 90 degrees clockwise, and -90 is 90 degrees counterclockwise, and so on. The range of valid values would be from -180 to +180.

PR #529, opened 2020-12-06
[nav-gimbal] Adds ZED data lcm file to receive current gimbal position from ESW

atguimond:
Resolves issue #528 

Yaw will be in degrees ranging from -180 to +180, with 0 degrees being straight ahead.

### DIFF
--- a/rover_msgs/ZedGimbalData.lcm
+++ b/rover_msgs/ZedGimbalData.lcm
@@ -1,0 +1,6 @@
+package rover_msgs;
+
+struct ZedGimbalCmd 
+{
+	double current_yaw;
+}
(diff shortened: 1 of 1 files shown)


rover_msgs/ZedGimbalData.lcm
```diff
@@ -0,0 +1,6 @@
+package rover_msgs;
+
+struct ZedGimbalCmd
```
jchen3652, line 3: This doesn't seem to match the filename - should this be ZedGimbalCmd?

jchen3652 (changes requested):

rover_msgs/ZedGimbalData.lcm
atguimond: Nope but the struct should be ZedGimbalData. James you are making a pull request for the ZedGimbalCmd, right?

amverni:
Discussed offline and this no longer meets our needs. The corrected version will be handled by #471

PR #530, opened 2020-12-06
[nav] Update Gate Traversal Algorithm

### LINKED ISSUE #340: Update Gate Traversal Algorithm
Replace the gate shimmy algorithm with algorithm that drives to far post as an approximation for centering rover between posts.

Have done some integration testing with this

zwgold:
resolves #340 

Replace GateShimmy code with code to drive the rover through the gate by having it drive to a post, then drive through the gate.

This change were tested by using the navigation simulator to mock several different rover scenarios when searching for a waypoint and turning/driving through the gate posts in the correct direction. This includes if the rover passed through the gate in the incorrect direction (and then reorient itself to pass through the correct direction).


amverni (changes requested):
Make sure to test this. I don't see any information about that.

jetson/nav/gate_search/gateStateMachine.hpp
```diff
@@ -53,7 +53,13 @@ class GateStateMachine
...
     NavState executeGateFace();
 
-    NavState executeGateShimmy();
+    NavState executeGateTurnToFarPost();
+
+    NavState executeGateDriveToFarPost();
+
+    NavState executeGateTurnToGateCenter();
+
+    //NavState executeGateShimmy();
```
amverni, line 62: ?

jetson/nav/rover.cpp
```diff
@@ -239,6 +239,7 @@ bool Rover::updateRover( RoverStatus newRoverStatus )
             mRoverStatus.obstacle() = newRoverStatus.obstacle();
             mRoverStatus.odometry() = newRoverStatus.odometry();
             mRoverStatus.target() = newRoverStatus.target();
+            mRoverStatus.target2() = newRoverStatus.target2();
```
amverni, line 242: Why `target` and `target2`? I believe they could be something more descriptive like `leftTarget` and `rightTarget`.

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -218,6 +218,7 @@ NavState SearchStateMachine::executeTurnToTarget( Rover* phoebe )
     if( phoebe->turn( phoebe->roverStatus().target().bearing +
                       phoebe->roverStatus().odometry().bearing_deg ) )
     {
+        cout << "Reached our target\n";
```
amverni, line 221: Remove debug print statements.

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -246,13 +247,38 @@ NavState SearchStateMachine::executeDriveToTarget( Rover* phoebe, const rapidjso
...
 
-    DriveStatus driveStatus = phoebe->drive( phoebe->roverStatus().target().distance,
+    if( phoebe->roverStatus().target2().distance > 0 )
+    {
+        if( phoebe->roverStatus().target().distance < phoebe->roverStatus().target2().distance ) 
+        {
+             driveStatus = phoebe->drive( phoebe->roverStatus().target().distance - 0.15,
+                                             phoebe->roverStatus().target().bearing +
+                                             phoebe->roverStatus().odometry().bearing_deg,
+                                             true );
```
amverni, line 260: Same comment as before.

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -246,13 +247,38 @@ NavState SearchStateMachine::executeDriveToTarget( Rover* phoebe, const rapidjso
...
+        }
+        else 
+        {
+             driveStatus = phoebe->drive( phoebe->roverStatus().target2().distance - 0.15,
+                                             phoebe->roverStatus().target2().bearing +
+                                             phoebe->roverStatus().odometry().bearing_deg,
+                                             true );
```
amverni, line 267: See previous.

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -246,13 +247,38 @@ NavState SearchStateMachine::executeDriveToTarget( Rover* phoebe, const rapidjso
...
+        }
+    }
+    else 
+    {
+        driveStatus = phoebe->drive( phoebe->roverStatus().target().distance - 0.15,
                                              phoebe->roverStatus().target().bearing +
                                              phoebe->roverStatus().odometry().bearing_deg,
                                              true );
```
amverni, line 275: See previous

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -246,13 +247,38 @@ NavState SearchStateMachine::executeDriveToTarget( Rover* phoebe, const rapidjso
         roverStateMachine->updateObstacleDistance( phoebe->roverStatus().obstacle().distance );
         return NavState::SearchTurnAroundObs;
     }
+    
+    DriveStatus driveStatus;
 
-    DriveStatus driveStatus = phoebe->drive( phoebe->roverStatus().target().distance,
+    if( phoebe->roverStatus().target2().distance > 0 )
+    {
+        if( phoebe->roverStatus().target().distance < phoebe->roverStatus().target2().distance )
```
amverni, line 255: This seems quite a bit more complicated than before. Let's add a high level comment before this block of code to describe what should be happening.

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -246,13 +247,38 @@ NavState SearchStateMachine::executeDriveToTarget( Rover* phoebe, const rapidjso
...
+        driveStatus = phoebe->drive( phoebe->roverStatus().target().distance - 0.15,
                                              phoebe->roverStatus().target().bearing +
                                              phoebe->roverStatus().odometry().bearing_deg,
                                              true );
+
+    }
+    
     if( driveStatus == DriveStatus::Arrived )
     {
+        cout << "Drove to target\n";
```
amverni, line 281: Remove debug prints

jetson/nav/stateMachine.cpp
```diff
@@ -218,8 +218,10 @@ void StateMachine::run()
             case NavState::GateTurnToCentPoint:
             case NavState::GateDriveToCentPoint:
             case NavState::GateFace:
-            case NavState::GateShimmy:
             case NavState::GateDriveThrough:
+            case NavState::GateTurnToFarPost: //new
+            case NavState::GateDriveToFarPost: //new
+            case NavState::GateTurnToGateCenter: //new
```
amverni, line 224: These comments don't seem to add much value and they are only temporarily accurate.

This goes for the code throughout.

amverni (changes requested):

jetson/nav/rover.cpp
zwgold: Because of how the code works, we don't need to differentiate between left and right targets. When we detect our first target, that is automatically listed as `target`, and if we can detect a second target, that is listed under `target2`. For instance, if we detected the right post first, that would go under `target`, or `leftTarget` with your suggestion (and when we detect the left post, that would go under `rightTarget` because of how our system detects them).

amverni: When you only see one target, it is the left (and the right) target. My point is that `target` and `target2` have little meaning in terms of adding context. Additionally, the second one spotted won’t always be `target2` just like left wouldn’t always stay as left (or right).

I was suggesting renaming for clarity not because you _needed_ to differentiate it.

jetson/nav/gate_search/gateStateMachine.cpp
```diff
@@ -72,7 +82,6 @@ NavState GateStateMachine::run()
     } // switch
 } // run
 
-//
```
amverni, line 75: Please make sure all functions here have a comment.

jetson/nav/gate_search/gateStateMachine.cpp
```diff
@@ -288,24 +294,24 @@ NavState GateStateMachine::executeGateDriveToFarPost()
...
 
+// Drive through gate posts
 NavState GateStateMachine::executeGateDriveThrough()
 {
     DriveStatus driveStatus = mPhoebe->drive( centerPoint2 );
 
     if( driveStatus == DriveStatus::Arrived )
     {
-        if(!CP1ToCP2CorrectDir)
+        if(!CP1ToCP2CorrectDir) // Check if we drove through the incorrect direction
```
amverni, line 314: Make spacing consistent with the rest of `jetson/nav` code. I.e., `if( !CP1ToCP2CorrectDir )`.

Please do this throughout.

jetson/nav/gate_search/gateStateMachine.cpp
```diff
@@ -167,10 +170,11 @@ NavState GateStateMachine::executeGateTurn()
     return NavState::GateTurn;
 } // executeGateTurn()
 
+// Drive to determined waypoint
 NavState GateStateMachine::executeGateDrive()
 {
-    if( mPhoebe->roverStatus().target2().distance >= 0 ||
-        ( mPhoebe->roverStatus().target().distance >= 0 && mPhoebe->roverStatus().target().id != lastKnownPost1.id ))
+    if( mPhoebe->roverStatus().rightTarget().distance >= 0 ||
+        ( mPhoebe->roverStatus().leftTarget().distance >= 0 && mPhoebe->roverStatus().leftTarget().id != lastKnownPost1.id ))
```
amverni, line 177: 1. `lastKnownPost1` should be something more like `lastKnownRightTarget` (consistency for right/left not 1 vs. 2 and consistency between calling it a target vs a post when we mean the same thing), right? Double check it's right, not left, please.
2. This line is really long. If it's over 100 (max 120), I would say make this two lines.
3. This is pretty tricky since you are checking both distances and IDs. Please add a comment explaining why.

jetson/nav/gate_search/gateStateMachine.cpp
```diff
@@ -225,53 +232,52 @@ NavState GateStateMachine::executeGateFace()
...
         {
             return NavState::GateDriveToFarPost;
         }
     }
-
     return NavState::GateTurnToFarPost;
 } // executeGateTurnToFarPost()
 
-// Drive to the farthest post, if possible
+// Drive to furthest post (or the only post if only one is available)
```
amverni, line 265: If we only see one post, isn't this the wrong state to be in? In that case, we would rather be looking for a second post, right?

jetson/nav/gate_search/gateStateMachine.cpp
amverni, line 314: `CP1ToCP2CorrectDir` seems like a poorly named variable (might be me to blame for that). Can you update it to be more readable? Even something simpler like `isCorrectGateDir`

jetson/nav/gate_search/gateStateMachine.cpp
amverni: You appear to be doing this a lot. Would it make sense to abstract this into a function?

jetson/nav/gate_search/gateStateMachine.cpp
```diff
@@ -327,27 +333,27 @@ NavState GateStateMachine::executeGateDriveThrough()
...
-                                          mPhoebe->roverStatus().target().bearing,
+                                          mPhoebe->roverStatus().leftTarget().bearing,
                                           360);
         lastKnownPost2.odom = createOdom( mPhoebe->roverStatus().odometry(),
                                           targetAbsAngle,
-                                          mPhoebe->roverStatus().target().distance,
+                                          mPhoebe->roverStatus().leftTarget().distance,
                                           mPhoebe );
-        lastKnownPost2.id = mPhoebe->roverStatus().target().id;
+        lastKnownPost2.id = mPhoebe->roverStatus().leftTarget().id;
```
amverni, line 356: Post2 -> LeftTarget

jetson/nav/gate_search/gateStateMachine.hpp
```diff
@@ -78,7 +78,7 @@ class GateStateMachine
     Odometry centerPoint1;
     Odometry centerPoint2;
 
-    //
+    // Check if we drove through the correct direction
     bool CP1ToCP2CorrectDir;
 
     double gateAdjustmentDist = 0.3;
```
amverni, line 84: This should be a constant.
1. use `const`
2. Should live somewhere more intuitive. I think the config file makes sense. However, another option would be the hpp file but I think the config file makes more sense.

Also, what is a gate adjustment distance?

jetson/nav/rover.cpp
```diff
@@ -50,16 +50,18 @@ Odometry& Rover::RoverStatus::odometry()
...
-Target& Rover::RoverStatus::target()
+Target& Rover::RoverStatus::leftTarget()
 {
     return mTarget1;
 } // target()
 
-Target& Rover::RoverStatus::target2() {
+Target& Rover::RoverStatus::rightTarget() 
+{
     return mTarget2;
```
amverni, line 60: `mTarget2` -> `mRightTarget` (same goes for left)

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -275,13 +275,13 @@ NavState SearchStateMachine::executeDriveToTarget( Rover* phoebe, const rapidjso
         {
             roverStateMachine->mGateStateMachine->mGateSearchPoints.clear();
             const double absAngle = mod(phoebe->roverStatus().odometry().bearing_deg +
```
amverni, line 277: spacing around parenthesis not consistent with rest of code

please do this throughout

jetson/nav/gate_search/gateStateMachine.cpp
```diff
@@ -218,74 +222,109 @@ NavState GateStateMachine::executeGateDriveToCentPoint()
...
+        }
     }
 
-    // If we need to switch directions
-    const bool visibleTargetAngles = mPhoebe->roverStatus().target().bearing > fovAngle / 2 &&
-                                     mPhoebe->roverStatus().target2().bearing < fovAngle / 2;
-    const bool visibleTargetDists = mPhoebe->roverStatus().target().distance < fovDepth &&
-                                    mPhoebe->roverStatus().target2().distance < fovDepth;
-    if(!visibleTargetAngles || !visibleTargetDists)
+    driveStatus = mPhoebe->drive( distance, bearing, true );
```
amverni, line 284: Why is `driveStatus` declared above? What about `DriveStatus driveStatus = mPhoebe->...` and then remove the declaration above.

jetson/nav/gate_search/gateStateMachine.cpp
```diff
@@ -218,74 +222,109 @@ NavState GateStateMachine::executeGateDriveToCentPoint()
...
-    const double gateAngle = calcBearing(lastKnownPost1.odom, lastKnownPost2.odom); // Angle from post 1 to post 2
-    const Odometry gateCent = createOdom(lastKnownPost1.odom, gateAngle, gateWidth / 2, mPhoebe);
-    const double roverToGateCentAngle = calcBearing(currOdom, gateCent); // ablsolute angle
-    mPhoebe->drive(direction, roverToGateCentAngle); // TODO: drive straight when going backwards
-    return NavState::GateShimmy;
-} // executeGateShimmy()
+// Execute turn back to center point for driving through the gate
+NavState GateStateMachine::executeGateTurnToGateCenter()
+{
+    if ( mPhoebe->turn( centerPoint2 ))
```
amverni, line 300: Remove space after `if`

jetson/nav/search/searchStateMachine.hpp
```diff
@@ -62,6 +62,8 @@ class SearchStateMachine {
     // Last known angle of rover from turn to target.
     double mTurnToTargetRoverAngle;
 
+    double searchAdjustmentDist = 0.15;
```
amverni, line 65: See my comment about `gateAdjustmentDist`; same applies here. Also, remember a comment where necessary.

amverni (changes requested):

amverni:
> Make sure to test this. I don't see any information about that.

The testing description you added using the simulator is perfect. However, you also need to do some integration testing (or, in extreme cases, at least creating an issue to track nav changes that need to be integration tested if no one from auton is able to do any integration testing for this in the near future). @mmartin4972 is a great person to talk to about what this might look like as he is working on this for his team.

jetson/nav/gate_search/gateStateMachine.cpp
zwgold: The lines with distance and ID checking were previously in there for the old gate drive code (not sure who had done that previously), so I kept it.

jetson/nav/gate_search/gateStateMachine.cpp
zwgold: I considered if there was only one post that we can see, that would be the "far" post we drive to (the code inside the executeGateDriveToFarPost function will determine if we only drive to a single post we see, or if we drive to a second one if it happens to be farther away). I can rename this if this is unclear

amverni: I think it's fine to leave the name as is. Make sure you update any documentation now that we have new states.

jetson/nav/gate_search/gateStateMachine.cpp
amverni: Got it. If you don't think that it's good to do this in your change, make an issue for it. If you do make a separate issue, make it more general than this. Something like "Clean up duplicated code" and in that have an item that is specifically these lines.

amverni:
@zwgold make sure you have all 3 types of reviewers laid out in the workflow wiki. It looks to me that you currently have 3 people but only 2 of the roles needed.

jetson/nav/gate_search/gateStateMachine.cpp
zwgold: Will do. I think it would be better to have a separate issue for it.

PR #545, opened 2020-12-19
[nav-sim] Add noise to simulated sensors

### LINKED ISSUE #316: Add optional noise to simulator
- [ ] Make noise for each piece of data simulated from other programs
  - [ ] `TargetList`
  - [ ] `Obstacle` 
  - [ ] `Odometry`
- [ ] Make button to turn on and off noise
  - [ ] `TargetList`
  - [ ] `Obstacle` 
  - [ ] `Odometry`

amverni:
resolves #316 

Add noise to the simulated localization and perception sensors. This noise can be turned on and off with the "Simulate Localization" and "Simulate Perception" buttons.

This change was tested manually by a series of tests with various combinations of ar tags and obstacles and "manual" driving of the rover in the simulator. Each test ensured that the noisy data made sense given the real data and that the real data was not modified by the noise.


simulators/nav/src/store/modules/roverState.ts
```diff
@@ -38,11 +52,32 @@ const state:RoverState = {
...
+    },
+    percepNoise: {
+      obsFalsePos: 0.1,
+      obsFalseNeg: 0.1,
+      tagFalsePos: 0.1,
+      tagFalseNeg: 0.1,
+      tagIdFalses: 0.1,
+      bearStddev: 5,
+      distStddev: 0.5
+    }
```
amverni, line 68: @mmartin4972 Can you help me choose reasonable values for these?

simulators/nav/src/store/modules/roverState.ts
```diff
@@ -38,11 +52,32 @@ const state:RoverState = {
     total_wps: 0
   },
 
+  noiseSetttings: {
+    locNoise: {
+      headingStdev: 5,
+      latLonStdev: 1
+    },
```
amverni, line 59: @rzf16 Can you help me choose reasonable values for these?

simulators/nav/src/utils/types.ts
```diff
@@ -163,17 +199,18 @@ export enum OdomFormat {
...
+   to Perception. */
+export interface PercepNoiseSettings {
+  obsFalsePos:number; /* percentage, [0, 1] */
+  obsFalseNeg:number; /* percentage, [0, 1] */
+  tagFalsePos:number; /* percentage, [0, 1] */
+  tagFalseNeg:number; /* percentage, [0, 1] */
+  tagIdFalses:number; /* percentage of tag IDs given incorrectly, [0, 1] */
+  bearStddev:number; /* degrees */
+  distStddev:number; /* meters */
+}
```
amverni, line 212: @mmartin4972

simulators/nav/src/utils/types.ts
```diff
@@ -121,6 +149,14 @@ export interface NavStatus {
 }
 
 
+/* Interface representing the settings that can be adjusted for the artificially
+   added noise. */
+export interface NoiseSettings {
+  locNoise:LocNoiseSettings;
+  percepNoise:PercepNoiseSettings;
+}
```
amverni, line 157: @rzf16

raytitan (commented):
Wouldn't "noisy perception" and "noisy localization" be more accurate? Isn't the simulator always "simulating" perception and localization?

amverni:
> Wouldn't "noisy perception" and "noisy localization" be more accurate? Isn't the simulator always "simulating" perception and localization?

Nope. Commonly forgotten fact: the navigation simulator can chose to turn off the simulating of perception or localization systems (or both). In fact, with a few more features, it could become a super powerful tool for use in integration testing which is something I was pushing for on the side (may be capable of it right now just not as friendly as it should be). Also, with a few more past that, it can also become the GUI because the aforementioned item is pretty close to just the GUI but that's not the point of this PR even though that would be sweet. So back to that...

That button rotates through simulating with noise, simulating without noise, and not simulating at all (I.e. taking input from the sensors). Other parts of the PR should highlight this. I agree that the naming isn't super clear but space is very valuable on this app and I didn't have a better name that was also concise. I'm open to ideas of different wordings though if you have any suggestions.

raytitan (dismissed):

simulators/nav/src/store/modules/roverState.ts
mmartin4972: We've never thrown a FP for tags, since our AR tag stuff is super picky. I'd set **tagFalsePos** to 0, or 0.01 if you still want to account for it. Since we're picky set **tagFalseNeg** to 0.15. Not sure if you account for us getting more accurate as we move closer to the waypoint, but on average I'd say 0.15 is probly a better representation. Set **obsFalseNeg** to 0.05. It's more likely that we see an obstacle that isn't actually there than don't see an obstacle that is there. 

We have ways to tune our code to give you more FP and fewer FN or vice versa, so after testing w the sim if you find you prefer more FN than FP or something like that lmk and we can tune our code accordingly

amverni: I adjusted the numbers to reflect your suggestions. I wanted to avoid 0 so I went with 1%. I attempted to implement getting more accurate as we got closer but it turned out that would be pretty complicated so I opted to put that off for now. That's definitely something that can be looked into in the future though.

atguimond (commented):
Are the noisy perception and localization the yellow buttons in the sim? I feel like this could be made more clear/user friendly.

amverni:
> Are the noisy perception and localization the yellow buttons in the sim? I feel like this could be made more clear/user friendly.

I'm very open to suggestions on this! I personally could not think of a way to do this without taking up more room on the UI which is already being pushed to the limit. I also don't think it's a huge deal since 1. it does make sense that the "Simulate Perception" button changes how and if we are simulating the perception sensors (and similar for localization) and 2. we can expect our users are experts in the subject matter of the autonomy system rather than designing our system for a general MRover member like `base_station/gui` has to be.

simulators/nav/src/store/modules/roverState.ts
rzf16: Sorry I'm getting to this late, I wasn't checking my email. What are the units you're using? For bearing, I think a standard deviation of ~5 degrees is safe. For latitude longitude, standard deviation of ~2 meters is reasonable.

raytitan (dismissed):

raytitan (approved):

rzf16 (approved):

zwgold (approved):

PR #547, opened 2020-12-19
[nav-sim] Make too close tags not visible

### LINKED ISSUE #543: Tag not visible when too close to it
It is not reasonable for an AR tag to be visible when we are right at it (I.e. 0 meters away). We should make a minimum distance away required to see the tag.

amverni:
resolves #543

This change makes tags within 0.25 meters be treated as
not visible in order to more closely mimic the ZED.

This change was tested by manual driving of the rover with
different tags at different distances and angles. It was also
verified that no interfaces were changed.


simulators/nav/src/utils/constants.ts
```diff
@@ -2,12 +2,19 @@
...
+  PerceptionConstants,
   PostConstants,
   RepeaterConstants,
   RoverConstants
 } from './types';
 
 
+/* Define the perception system constants */
+export const PERCEPTION:PerceptionConstants = {
+  minVisibleDistTag: 0.25
```
amverni, line 14: @mmartin4972 can you verify this number makes sense?

raytitan (approved):

simulators/nav/src/utils/constants.ts
mmartin4972: I tested it out and if the tag is 15cm the top and bottom borders get cutoff, so I think 0.25 is a good value here

atguimond: Do we have any idea how close the camera will be to the plane of the AR tag or does URC make that a variable height? I am just thinking if the angle between the camera and the tag is too great then maybe we would need a larger margin of error.

amverni: The tennis balls used to be about a meter off the ground, but we've never seen the AR tags at comp for obvious reasons. This number can be adjusted pretty easily in the future if we find that it's way off the mark. Additionally, you can temporarily change it locally to test with different amounts and see how the nav system reacts. In fact, you bring up a good point in that this might be worth doing in testing.

atguimond: Okay that sounds good. We should put it on the agenda to test during integration testing to get an idea of the number.

atguimond (approved):

simulators/nav/src/utils/constants.ts
mmartin4972: Sounds good

mmartin4972 (approved):

Issue #552, opened 2020-12-21
Fix Silent Mode


mmartin4972:
Currently our code doesn't compile with perception_debug=false. We need to fix these build issues and configure this build flag, so when it is set to false their is no output from the program whatsoever to help maximize performance.

mmartin4972:
Fix line 50 TODO in test_protocol.md once this mode has been made active

mmartin4972:
Fix line 37 in test_protocol.md once this mode has been made active

PR #573, opened 2021-01-17
[percep] Added console display 

MarcSocha:
Added console display displaying specific stats about various variables. Used for debugging. Added Display class for display.

## Compile Test
- [ ] Offline Debug: ` ./jarvis build jetson/percep -o with_zed=false ar_detection=true obs_detection=true perception_debug=true `
  - [ ] No compiler errors or warnings
- [ ] Offline Silent: ` ./jarvis build jetson/percep -o with_zed=false ar_detection=true obs_detection=true perception_debug=false `
  - [ ] No compiler errors or warnings
- [ ] ZED Debug: ` ./jarvis build jetson/percep -o with_zed=true ar_detection=true obs_detection=true perception_debug=true `
  - [ ] No compiler errors, only ZED warnings
- [ ] ZED Silent: ` ./jarvis build jetson/percep -o with_zed=true ar_detection=true obs_detection=true perception_debug=false `
   - [ ] No compiler errors, only ZED warnings

## Offline Debug Test
- [ ] `./jarvis build jetson/percep -o with_zed=false ar_detection=true obs_detection=true perception_debug=true`
- [ ] Run code against test1 image set located in perception-files repository
- Check for:
   - [ ]  Left and right path are being calculated
   - [ ]  Calculated bearing is reasonable given viewer output
   - [ ]  AR Tag ID is 2
   - [ ]  Obstacles with large width are detected
   - [ ]  Path is projected around not through large obstacles
   - [ ]  Short obstacles are detected
   - [ ]  Steep inclines are detected as obstacles
- [ ] Run code against test2 image set located in perception-files repository
- Check for:
   - [ ]  Left and right path are being calculated
   - [ ]  Calculated bearing is reasonable given viewer output
   - [ ]  AR Tag ID is 6
   - [ ]  2 AR Tags detected simultaneously
   - [ ]  Obstacles with large width are detected
   - [ ]  Path is projected around not through large obstacles
   - [ ]  Short obstacles are detected
   - [ ]  Steep inclines are detected as obstacles

## Offline Silent Test
- [ ] `./jarvis build jetson/percep -o with_zed=false ar_detection=true obs_detection=true perception_debug=false`
- [ ] Run `time ./jarvis exec jetson/percep` against test_case1 image set
- [ ] Run `time ./jarvis exec jetson/percep` against test_case2 image set
- [ ] User value is less than or equal to \<TODO Determine Value\>

## Unit Test
- [ ] ` ./jarvis build jetson/percep -o with_zed=true ar_detection=true obs_detection=true perception_debug=true `
- [ ] `./jarvis exec lcm_tools_echo Obstacle "/obstacle"`
- [ ] `./jarvis exec lcm_tools_echo TargetList "/target_list"`
- Test environment contains:
  - [ ] 1 large obstacle ~ over 40cm tall
  - [ ] 2 small obstacles ~  25cm-30cm tall
  - [ ] 1 post
  - [ ] 1 hill
- Test for:
  - [ ] Detected large obstacle 6m away
  - [ ] Detected small obstacle 2m away
  - [ ] Detected clear path when 1.2m gap between 1 large and 1 small obstacle
  - [ ] Left and right path are being calculated
  - [ ] Calculated bearing is reasonable given viewer output of path
  - [ ] AR Tag ID is the correct ID
  - [ ] Obstacles with large width are detected
  - [ ] Short obstacles are detected
  - [ ] FPS values don't exceed \<TODO find this value in Issue #552\>
  - [ ] Obstacle.lcm in /obstacle is populated correctly
  - [ ] TargetList.lcm in /target_list is populated correctly

## Integration Test
- [ ] Create new branch based off of main
- [ ] Checkout new branch
- [ ] Cherry-Pick all PRs that need testing onto new branch
- [ ] Compile all Software Packages
- [ ] Run all code simultaneously
- [ ] Obstacle.lcm in /obstacle is populated correctly
- [ ] TargetList.lcm in /target_list is populated correctly
- [ ] Sensors are Active
  - [ ] Stereo Camera (ZED)
  - [ ] Inertial Measurement Unit (IMU)
  - [ ] Global Positioning System (GPS)


jetson/percep/display.hpp
```diff
@@ -0,0 +1,56 @@
...
+using namespace std;
+
+class Display{
+  private:
+  Mat img{Mat(250, 400, CV_8UC3, Scalar(0,0,0))};
+  string windowName;
+  map<string, double> inputStats;
+
+
+  void clearDisplay(){
```
mmartin4972, line 14: Please implement all these functions in a .cpp file

jetson/percep/main.cpp
```diff
@@ -72,9 +76,17 @@ int main() {
   cam.record_ar_init();
   #endif
 
-
+  Display display("Console Display");
 /* --- Main Processing Stuff --- */
   while (true) {
+    #if PERCEPTION_DEBUG
+      auto fpsStart = std::chrono::high_resolution_clock::now();
```
mmartin4972, line 83: Ideally we don't want to have a bunch of high resolution clock calls throughout our code, since this can make things quite messy. Ashwin has written a [timer class](https://github.com/umrover/perception-gpu-obs-detector/blob/main/include/Timer.hpp) that might be helpful here to integrate.

mmartin4972 (changes requested):

mmartin4972:
My vision for this class was to be a more general debug class as opposed to specifically a display class. The idea would be that if we ever wanted to print something out we could just call `display.insert({<string of text defining output>, the type of output, the actual value of output})`. I was thinking that we could then take advantage of the type of output function, so that if we are trying to display a time we could call `display.insert({<name of timer>}, enum::time, <enum::current or enum::duration>`. If it is the current enum it would add the start time to the hash map and if it is the duration enum it will do the necessary math to find the duration duration based on the current time and the time that is currently located in the hash map. If the command is   `display.insert({<name of timer>}, enum::text, value to display` then it would just add the name and value to the map. When `display.show` is called at the end we would add all the elements in the map to the display and print it out. In summary, this display would have an insert function that is able to accept parameters specifying whether it wants timing output or just generic text output and then display it all at the end. This object would be held in local scope and could be added to by any program. If we want to add the information from the map to a filestream at some point we could easily add this functionality, just by modifying the display class itself as opposed to modifying where all of the display functions were called throughout the code. The only related code that we'd then see throughout the main code files would be the constructor,`display.insert()`, and `display.show()` functions. This will make sure they don't make our code super messy and are simple and easy to use. I am going to close this PR until we have implemented these adjustments

PR #575, opened 2021-01-22
[nav] Update Gate Traversal Algorithm

### LINKED ISSUE #340: Update Gate Traversal Algorithm

zwgold:
resolves #340

Replace GateShimmy code with code to drive the rover through the gate by having it drive to a post, then drive through the gate.

This PR was originally on main. However, we have closed the original review and moved it to the auton integration testing branch so we can run it before merging it into the main repository.

These changes were tested by using the navigation simulator to mock several different rover scenarios when searching for a waypoint and turning/driving through the gate posts in the correct direction. This includes if the rover passed through the gate in the incorrect direction (and then reorient itself to pass through the correct direction).


aswarner:
Good style throughout. I tested this on the sim and it worked aside from the occasional "driving off into the sunset" phenomenon that we attribute to the sim not resetting. I tested approaching the gate forwards and backwards. So excited to see this on the rover!

aswarner:
We talked about extending the "center point" distances from the gate after real testing, to allow the rover more distance to cross if necessary.

aswarner (approved):
Good style throughout. I tested this on the sim and it worked aside from the occasional "driving off into the sunset" phenomenon that we attribute to the sim not resetting. I tested approaching the gate forwards and backwards. So excited to see this on the rover!

raytitan (approved):

so306:
Nice formatting/spacing, good renaming calls (ex. Post1/2 changed to RightPost/LeftPost). I tested on sim to the best of my ability! Seemed to work.

so306 (approved):

PR #577, opened 2021-01-24
[gui] Auton GUI Redesign, raw data, waypoint editor fixes etc.

jnnanni:
Added raw data LCMs in the scrollable area under the camera box, fix a bug where adding too many waypoints to the waypoint editor would squish the boxes above it, moved top right odom reading to be inline with the other data in that area.

### DIFF
--- a/base_station/gui/src/components/AutonJoystickReading.vue
+++ b/base_station/gui/src/components/AutonJoystickReading.vue
@@ -1,7 +1,7 @@
 <template>
 <div class = "main-content">
  <h3>Joystick Reading</h3>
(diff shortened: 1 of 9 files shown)


raytitan (commented):
Mind if we merge this into Auton_Integration_Test first? That way the auton team will be able to try this out

skorman (approved):

ErikLytle (approved):

DavidSmith166 (approved):

raytitan (approved):
Let's see how well this works

PR #581, opened 2021-01-28
Revert "[nav-sim] Add noise to simulated sensors"

aswarner:
Reverts umrover/mrover-workspace#545
Once merged, we noticed that the noise setting was too noisy to use and the no-noise setting introduced bugs like incorrect detection of posts
Reverting now rather than waiting for bug correction so that a functional sim is maintained in main branch


ankithu (approved):

raytitan (commented):
Has someone checked out the branch created by this revert PR, and tested the simulator?

aswarner:
@raytitan yes, Zack just confirmed that the reverted sim works

raytitan (approved):

DavidSmith166 (approved):

PR #586, opened 2021-02-05
[Nav/ZedGimbal] Adds Zed Gimbal Functionality to Search and Gate Traversal State Machines

ankithu:
#388 , #389 

This commit adds a Gimbal class to allow control of the Zed Gimbal for improved search efficiency. It also adds the use of that gimbal class through added states to both the Gate Traversal and Search state machines. The basic functionality is that instead of turning 360 degrees when we get to a point we want to search. We turn the gimbal to positive 150 degrees then to -150 degrees then back to 0. This number is editable through the config. Additionally, code was added to receive and send gimbal LCM messages. Additionally, as this is a new feature, a boolean was added to the config that can be turned on or off to enable or disable Zed Gimbal functionality.

Changes were tested in the simulator. The search state machine changes were tested with various locations of waypoints and AR tags, in all instances that I tested, the rover was able to successfully find the AR tags. The gate traversal algorithm also worked with various orientations of the gates and distances from the point (enough to enter the diamond search pattern) and the rover made it through even with the gate pointing in a direction that would be opposite from ideal. Additionally, the enable and disable gimbal switch was tested and works in the sim.


raytitan (commented):
This goes in auton_integration_test branch first right?

aswarner:
@raytitan yes, that's my fault. will close and re-open

PR #606, opened 2021-02-22
SAR-critical improvements to allow rover to reach targets when posts are detected as obstacles

aswarner:
Added function isObstaclePost() in utilities.cpp, called by isTargetReachable(). isTargetReachable() now returns true if the obstacle does not block the target OR if the detected obstacle is within {configurable, currently 1.5} meters of the detected tag

### DIFF
--- a/config/nav/config.json
+++ b/config/nav/config.json
@@ -31,6 +31,7 @@
 		"drivingBearing": 50,
 		"waypointDistance": 2.0,
 		"targetDistance": 1.0,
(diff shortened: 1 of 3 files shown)


kayjohn1512 (approved):
Looks good

raytitan (commented):
You should fix this
https://travis-ci.org/github/umrover/mrover-workspace/builds/760047217#L13791

PR #609, opened 2021-02-25
[teleop] Turn off motors when auton is killed

skorman:
When creating this code we made a branch off of main. Seems like it cannot merge due to changes in auton_integration that are not on main. The only file changed is jetson/teleop/src/__main__.py so I believe we should be able to merge by keeping everything else the same from auton_integration


aswarner:
@skorman can you resolve the merge conflicts?

aswarner:
@skorman can you resolve the merge conflicts?

aswarner (changes requested):
there are merge conflicts in some files that need to be resolved

aswarner (commented):
sorry im dumb

aswarner:
closing to make a pr for just the file we need

PR #610, opened 2021-02-25
[teleop] turn off motors when nav is killed from command line

aswarner:
added teleop's new src/__main__.py which makes sure motors stop when nav is killed with ctrl C in command line


aswarner:
i messed up we only want one file changed

PR #611, opened 2021-02-25
[teleop] turn off motors when nav is killed from command line

aswarner:
added teleop's file for stopping motors


skorman (approved):
I discussed this with Ana but to everyone else, this program will send a kill command regardless of whether auton was terminated from the command line or not. I believe this should be okay, but just wanted to let everyone else know.

PR #615, opened 2021-03-07
[nav] Added configurable obstacle distance threshold function

### LINKED ISSUE #499: [Auton-Nav] Define a obstacle threshold in statemachine
- Nav should determine when it is necessary to move around an obstacle
- This should be easily configurable at comp

radraju:
Resolves #499

Added a function (isObstacleInThreshold) to check if an obstacle’s distance from the rover is less than or equal to the distance threshold defined in the nav config file. This function runs whenever we check if an obstacle is detected, so the rover only enters obstacle avoidance state when it is within the distance threshold.

This change was tested in the navigation simulator to make sure obstacle avoidance worked as usual and was only entered when the rover was within a specific distance threshold.


DavidSmith166 (dismissed):

so306 (dismissed):

aswarner (approved):
Great work!

so306 (approved):

DavidSmith166 (approved):

PR #616, opened 2021-03-07
[nav] fix sim downloading json files

aswarner:
fixed a bug that meant test case json files weren't downloading


raytitan:
Looks like the travis build is failing on the snafoo project in auton integration. I'll remove it and restart the check

PR #626, opened 2021-03-18
[Nav/gimbal-turn-to-target] Adds Gimbal Based Target Verification Logic

### LINKED ISSUE #592: [Auton-Nav] Improved Zed Gimbal Functionality
Use the Zed gimbal to add improved target confirmation and better target loss functionallity. Basically, if we find a target, turn the gimbal towards it in order to make sure we are actually seeing it. If we lose a target, turn the gimbal towards the last known target to try to reestablish the target for CV.

ankithu:
Resolves #592

This PR adds code to verify targets using the gimbal. Basically, whenever we see a target, before potentially wasting energy and wearing out the motors by actually turning the rover towards it, we first turn the gimbal towards the target and then hold it for a certain amount of time to verify that the target has been actually acquired, then once it has been confirmed, we turn the rover towards the target, move the gimbal back to 0 then move onto the next searchstate.

Code was tested in the simulator by running at multiple different angles and positions and testing the full valid cycle along with testing while deleting the AR-tag while the gimbal is turning or verifying its existence.

### DIFF
--- a/jetson/nav/search/searchStateMachine.cpp
+++ b/jetson/nav/search/searchStateMachine.cpp
@@ -287,7 +287,7 @@ NavState SearchStateMachine::executeSearchDrive( Rover* phoebe, const rapidjson:
 // Executes the logic for turning to the target.
 // Starts by turnign the gimbal towards the target. Then holds for 200 cycles to confirm aquisition
 // If the rover loses the target, will continue to turn using last known angles.
-// Once target location confirmed, rover turns towards the target, then rights its gimbal
+// Once target location confirmed, rover turns towards the target, then re-centers its gimbal
 // If the rover finishes turning to the target, it goes into waiting state to
 // give CV time to relocate the target
 // Else the rover continues to turn to to the target.
@@ -298,13 +298,15 @@ NavState SearchStateMachine::executeTurnToTarget( Rover* phoebe, const rapidjson
 
     static bool firstTime = true;
     
-    //if we are not using a gimbal, just skip all gimbal states and immeaditly turn to target, only needs to run once
+    //if we are not using a gimbal, just skip all gimbal states and immediately turn to target, only needs to run once
     if ( !roverConfig[ "search" ][ "useGimbal" ].GetBool() && firstTime){
         turnToTargetStage = 5;
         firstTime = false;
         bearingAngle = phoebe->roverStatus().leftTarget().bearing + phoebe->roverStatus().odometry().bearing_deg;
     }
     
+    //if we lose the target and we are not in a state where this is expected (re-centering gimbal, turning to target (gimbal might go off))
+    //then notify loss of target, reset static vars, and turn back to last known angle on the search path
     if( phoebe->roverStatus().leftTarget().distance < 0  && turnToTargetStage != 2 && turnToTargetStage != 3)
     {
         cerr << "Lost the target. Continuing to turn to last known angle\n";
@@ -318,6 +320,8 @@ NavState SearchStateMachine::executeTurnToTarget( Rover* phoebe, const rapidjson
         return NavState::TurnToTarget;
     }
    
+    
+    //first step (with gimbal): turn gimbal to the target with a margin of error of 10 degrees
     if ( turnToTargetStage == 0 ){
        
         if ( abs( phoebe->roverStatus().leftTarget().bearing - phoebe->gimbal().getYaw() ) > 10 )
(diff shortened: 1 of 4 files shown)


jetson/nav/search/searchStateMachine.cpp
```diff
@@ -284,28 +285,86 @@ NavState SearchStateMachine::executeSearchDrive( Rover* phoebe, const rapidjson:
 } // executeSearchDrive()
 
 // Executes the logic for turning to the target.
+// Starts by turnign the gimbal towards the target. Then holds for 200 cycles to confirm aquisition
```
aswarner, line 288: Instead of 200 cycles, this could be done using the perception wait time defined in the config, it would help with standardization and using meaningful units

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -284,28 +285,86 @@ NavState SearchStateMachine::executeSearchDrive( Rover* phoebe, const rapidjson:
...
 // If the rover loses the target, will continue to turn using last known angles.
+// Once target location confirmed, rover turns towards the target, then rights its gimbal
 // If the rover finishes turning to the target, it goes into waiting state to
 // give CV time to relocate the target
 // Else the rover continues to turn to to the target.
-NavState SearchStateMachine::executeTurnToTarget( Rover* phoebe )
+NavState SearchStateMachine::executeTurnToTarget( Rover* phoebe, const rapidjson::Document& roverConfig  )
 {
-    if( phoebe->roverStatus().leftTarget().distance < 0 )
+    static int turn_to_target_state = 0;
```
aswarner, line 296: change name to stage instead of state. also, use camelCase throughout instead of snake case

aswarner (changes requested):

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -284,28 +285,92 @@ NavState SearchStateMachine::executeSearchDrive( Rover* phoebe, const rapidjson:
 } // executeSearchDrive()
 
 // Executes the logic for turning to the target.
+// Starts by turnign the gimbal towards the target. Then holds for 200 cycles to confirm aquisition
```
aswarner, line 288: can you update this comment to reflect the use of a time wait?

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -284,28 +285,92 @@ NavState SearchStateMachine::executeSearchDrive( Rover* phoebe, const rapidjson:
...
+            firstTime = true;
             return NavState::TurnedToTargetWait;
         }
         return NavState::TurnToTarget;
     }
-    if( phoebe->turn( phoebe->roverStatus().leftTarget().bearing +
-                      phoebe->roverStatus().odometry().bearing_deg ) )
-    {
+   
+    if ( turnToTargetStage == 0 ){
```
aswarner, line 321: can you comment what each stage does? just one line for each is fine. I see it's kinda outlined in the function description, but specificity in the implementation here would help

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -284,28 +285,92 @@ NavState SearchStateMachine::executeSearchDrive( Rover* phoebe, const rapidjson:
...
+            
+            timeStarted = true;
+        }
+        double waitTime = roverConfig[ "search" ][ "searchWaitTime" ].GetDouble();
+        if ( difftime( time( nullptr ), startTime ) > waitTime ){
+            timeStarted = false;
+            ++turnToTargetStage;
+        }
+    }
+    else if (turnToTargetStage == 2){
```
aswarner, line 347: spaces inside ( ) for else if states 2,3, and 4

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -284,28 +285,92 @@ NavState SearchStateMachine::executeSearchDrive( Rover* phoebe, const rapidjson:
...
+        
+        if( phoebe->turn( bearingAngle ) )
```
aswarner, line 349: fix rover not turning to target and turn to target wait when gimbal=false

aswarner (changes requested):
buggy behavior for rover turning to targets without gimbal - switches to unexpected states instead of turning to last location and waiting

aswarner (changes requested):
buggy behavior: rover gets stuck in turn to target state

cgiger00:
update: will come back to at the start of the year

PR #627, opened 2021-03-18
[Auton/Nav] Fixed sim obstacle bug

### LINKED ISSUE #584: [Auton-Nav] Make Simulator report Posts as Obstacles
We found that any target that the rover might encounter on its way to a waypoint is not being interpreted as an obstacle. This has resulted in a bug where the rover can pass/drive through a target without harm to travel to the waypoint. This is a simulator issue.

rnagpal02:
resolves #584

Fixed nav simulator bug where AR tags and gates were not being detected as obstacles. Changed code so whenever AR tag or gate is changed, `onObstaclesChange` is called where an obstacle for every AR tag and post is also added to `obstacle_detector`'s array of obstacles. This makes it so the extra added obstacles are not visible on the UI, but perception still detects them.

Code was tested by running sim and placing AR tags and gates between waypoints and making sure rover avoided them. Deletion functionality was also tested by adding a tag or gate and then deleting it and making sure the rover would not try to drive around it. This was done using individual deletes and the clear field option.

### DIFF
--- a/simulators/nav/src/components/perception/Perception.vue
+++ b/simulators/nav/src/components/perception/Perception.vue
@@ -13,6 +13,7 @@
 <script lang="ts">
 import { Component, Vue, Watch } from 'vue-property-decorator';
 import { Getter, Mutation } from 'vuex-class';
+import { calcRelativeOdom } from '../../utils/utils';
 import {
   ArTag,
   FieldOfViewOptions,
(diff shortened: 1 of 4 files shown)


aswarner (approved):
This looks awesome! You implemented the posts as obstacles very elegantly, great job! Also good style and commenting.

PR #646, opened 2021-04-01
[percep] Added second bearing to Obstacle LCM

vidyasilai:
Added a right bearing variable to the Obstacle LCM struct, in order to let nav decide what clear path to take. Also added left and right bearing variables to the PCL class and obstacle_return classes.


mmartin4972:
Seems to have been opened by mistake

PR #647, opened 2021-04-04
Consolidated Packages Installed by Ansible 

### LINKED ISSUE #598: Consolidate packages installed by multiple ansible scripts
Currently, we have jetson, base_station, and cv roles in our [ansible folder](https://github.com/umrover/mrover-workspace/tree/main/ansible), which install many redundant packages. The devbox role calls on each of these roles as well.

The ideal solution is probably to have one "common" role, to install packages shared by most roles, and have those roles call on the "common" role. However it should be researched whether this is the ideal solution.

This will require lots of testing - using ansible-playbook quite often. Also, to have true testing, the virtual machine should be started from fresh for each test.

aidandeli:
Resolves #598

These changes eliminate any repeat packages installed by ansible when calling certain playbooks. Any packages that were shared between certain ansible tasks called in the same playbook were consolidated to a common_dev task.  This does install unnecessary packages for when certain playbooks are run.

This was tested by running each playbook with the common_dev task on a new VM each time. This was also tested by doing jarvis build all on a clean VM as well. NOTE I was only able to build 13/39 projects, but this was the case for both the code straight from the mrover-workspace and my local fork (w/ common_dev and changes).

### DIFF
--- a/ansible/base_station.yml
+++ b/ansible/base_station.yml
@@ -9,6 +9,7 @@
   roles:
   - jetson_dev
   - base_station_dev
+  - common_dev
   - base_station_networks
   - base_station_services
   tasks:
--- a/ansible/devbox.yml
+++ b/ansible/devbox.yml
@@ -9,3 +9,4 @@
   - base_station_dev
   - cv_dev
   - pis_dev
+  - common_dev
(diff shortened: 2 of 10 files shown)


cgiger00:
closing this for now @CameronTressler

PR #650, opened 2021-04-08
[Teleop] Refactor kinematics to C++

CameronTressler:
Complete refactoring of kinematics to C++ in the ra_kinematics package.


raytitan:
1. Please name this properly

base_station/gui/src/components/Ammonia.vue
```diff
@@ -0,0 +1,46 @@
+<template>
```
raytitan, line 1: There shouldn't be science GUI changes in here should there?

.gitignore
```diff
@@ -13,3 +13,4 @@ build/
 output.log
 base_station/gui/src/static/map
 /core.*
+jetson/ra_kinematics/meson.build
```
raytitan, line 16: why is this in the .gitignore

base_station/gui/src/components/AutonJoystickReading.vue
```diff
@@ -1,7 +1,7 @@
 <template>
```
raytitan, line 1: why is this in here?

base_station/gui/src/components/AutonTask.vue
```diff
@@ -26,25 +26,30 @@
     <div class="box1 data">
```
raytitan, line 26: why is this in here?

raytitan (changes requested):
There seem to be 300 files in here that are not related to ra_kinematics. Please use an interactive rebase to drop any commits not in main and not related to ra_kinematics, or remake this PR with a branch that only diffs from main with the jetson/ra_kinematics folder

PR #657, opened 2021-04-09
[Nav] Gimbal: Removes redundant gimbal call

### LINKED ISSUE #656: [Auton/Nav] Fix Gimbal Interface
Right now you need to call rover->gimbal.setDesiredYaw() and then also call rover->publishGimbal(); this is redundant, and rover->publishGimbal should not need to be called.

ankithu:
Fixes #656 

This PR fixes the problem of having to call two separate functions to move the gimbal. Previously you had to get the gimbal object from the rover then call the set desired yaw on it. And because certain objects required for sending LCM messages were only in the scope of rover.cpp, we had another method in rover.cpp to actually send that information to the gimbal class so it could publish the message

This PR creates a new method in the rover class that handles both the setting of gimbal yaw and publishing the LCM command (both through calling functions in the gimbal class). This prevents the user from having to access the gimbal object directly and also prevents the user from having to make two function calls.

This code has been tested in the simulator

### DIFF
--- a/config/nav/config.json
+++ b/config/nav/config.json
@@ -78,7 +78,7 @@
 		"searchWaitStepSize": 90.0,
 		"gimbalSearchWaitStepSize": 75.0,
 		"searchWaitTime": 1.0,
-		"useGimbal": false,
+		"useGimbal": true,
 		"gimbalSearchAngleMag": 150
 	}
 }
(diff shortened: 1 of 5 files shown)


rnagpal02 (approved):
Is `bool Rover::sendGimbalSetpoint(double desired_yaw)` replacing `setDesiredGimbalYaw` as the new public interface for controlling the Gimbal?

PR #658, opened 2021-04-11
[percep] Added launch capabilities to jarvis

mmartin4972:
The addition of the launch command allows users to build and execute an entire subsystem using the `./jarvis launch auton` command. The command will open up each subteam's programs in a new terminal and creates new tabs within the terminal and execs the multiple commands associated with each subteam in these tabs. @aswarner please check out the nav file and the launch_nav function to make sure I am building and running everything correctly. @rzf16 please do the same thing for the loc file and the launch_loc function. 

Before this gets merged in I'd like to test it on a jetson to make sure it works over ssh. Obviously want to test at an integration test as well before merging in. 

No other jarvis commands should be limited by the addition of this command


mmartin4972:
This needs to go into auton_integration before going into main

PR #659, opened 2021-04-11
[percep] Added CODEOWNERS file for perception

mmartin4972:
Code Owners is a type of branch protection that is provided by github. It requires that any PRs that change files listed in the CODEOWNERS doc can only be changed with an approving review of the person or one of the persons associated with those files. Specifically from a Perception perspective, if anyone wants to PR code into main that would affect the files listed in the CODEOWNERS file they would have to add mmartin4972 as one of the 3 reviewers needed for the PR to go through. 

The intent of this is more geared towards the auton_integration branch than main, since it will prevent subteams from accidentally merging in code that affects another subteam. This is far more common in auton_integration since most teams using auton_integration have their code in a different state relative to main and are trying to use auton_integration to test a PR that is going into main. However, when creating a new PR that PRs their branch that was previously going into main, into auton_integration git may try to make changes to other sub-teams' code, since it is trying to make their code the same as the code associated with their subteam in main. I put together this [doc](https://docs.google.com/document/d/1c6ynwNDP1vwB-k5Rd9IqQrspph9a1DJHOpMwDtqnXZo/edit?usp=sharing) that goes over how to PR commits into auton_integration given you have a branch that is ready to be PR'd into main. Here is more [documentation](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners) on the specifics of Code Owners.


cgiger00:
@mmartin4972

cgiger00 (approved):

PR #693, opened 2021-07-25
Float4fixmerge

mmartin4972:
Refactored GPU code to use the cuda float4 instead of the sl::float4. Also cleaned Ransac code


niwhsa9 (approved):
this is awesome. filter.hpp/cu is a really nice design. 

Left a couple of minor nit-picks that can either be fixed now or made into TODO comments so we remember later.

PR #751, opened 2021-10-03
[teleop] Complete Redesign of Science GUI

jnnanni:
All of the changes to the Science GUI made during the last school year. Changes are teleop-science-gui-improvments project. The changes in this PR are in the in progress column.


CameronTressler:
Instead of creating a Google Doc with screenshots of LCMs, can you add a README.md file to base_station/gui/ and enumerate which LCM messages/channels are supposed to be sent and received (and say for each that you tested it). For now it can be specific to the Science GUI even though it's in the gui/ directory

CameronTressler (approved):
Readme will need to be updated with more detail in the future, but I think this should be sufficient for now

base_station/gui/src/components/SATask.vue
```diff
@@ -49,11 +58,15 @@ import { mapGetters, mapMutations } from 'vuex'
 import Cameras from './Cameras.vue'
 import RoverMap from './RoverMap.vue'
 import CommIndicator from './CommIndicator.vue'
-import OdometryReading from './OdometryReading.vue'
+import OdometryReading from './OdometryReadingSA.vue'
```
cgiger00, line 61: do we need odom data in science?

cgiger00: we aren't planning for the gps to be plugged in at least during this

base_station/gui/src/static/rover_msgs.json
```diff
@@ -890,6 +1018,7 @@
         "MicroCam",
         "Microscope",
         "Mosfet",
```
cgiger00, line 1020: can you get rid of this

cgiger00 (changes requested):

PR #789, opened 2021-11-16
Added visualizer rotation around an arbitrary target

### LINKED ISSUE #773: Fix Visualizer Rotation
The visualizer currently allows us to rotate the camera around the point cloud to view it from all angles, but it is centered at 0,0,0 and the point cloud is not. Lets compute the centroid of the point cloud and center our rotation around that.

JasonLiu2002:
Fixes #773

### DIFF
--- a/jetson/percep_obs_detect/include/viewer.hpp
+++ b/jetson/percep_obs_detect/include/viewer.hpp
@@ -88,32 +88,42 @@ class Object3D {
 // Defines the view 
 class FirstPersonCamera {}; // TODO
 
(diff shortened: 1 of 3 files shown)


niwhsa9 (approved):
nice, before you squash and merge, can you add this below line 79 of obs_dector.cu?
`        
if(frameNum == 1) viewer.setTarget();
`
So it will look like:
```
...
else if(source == DataSource::FILESYSTEM) {

        pc = fileReader.readCloudGPU(frameNum);
        if(frameNum == 1) viewer.setTarget();

    }
...
```

PR #797, opened 2021-11-30
[travis] Added dockerfile for development container

mmartin4972:
Created a new dockerfile min.Dockerfile which will construct a minimal docker container for developing rover code for non-Perception team members. The container should be far faster to construct and take less space than percep.Dockerfile.

These dockerfiles are not actively used for being constructed into images. 
The percep.Dockerfile image can be found here: https://hub.docker.com/repository/docker/umrover1/travis. 
The min.Dockerfile image can be found here:  https://hub.docker.com/r/mmartin4972/rover. 

Currently there are no scripts for constructing these images and updating the images must be done manually.

The addition of these files has no effect on the existing code in the repository. They are just there for potential future modification and update.


CameronTressler (approved):

niwhsa9 (approved):
LGTM. 

On a side note: once we get GPU Obs Detection into main the percep dockerfile shouldn't look all that different from this minimal one. Only additions are GLM, CUDA, and ZED SDK. But agreed that keeping two is better anyway.

zwgold (approved):

mmartin4972:
Perception has developed an updated development container. Closing this PR so that they can add their new Dockerfile in another PR.

PR #799, opened 2021-11-30
[jarvis] Fixed failing build dependency of lcm_tools/send

mmartin4972:
Added the rover_msgs dependency to lcm_tools/send, which was causing its build to fail, and consequently the travis check to fail. 

Normally lcm_tools/common would build before lcm_tools/send and lcm_tools/common has a dependency on rover_msgs, so it would build rover_msgs. However, lcm_tools/common started building after lcm_tools/send for some reason and we got the build bug. 

Started pulling the docker image from umrover1 dockerhub user instead of umrover user because we don't have the login credentials for the umrover user and cannot recover them

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: required
 services:
 - docker
 script:
-- docker pull umrover/travis
+- docker pull umrover1/travis
 - docker run --rm -it -v $(pwd):/root -v /dev/tty:/dev/ttyUSB0 -v /dev/null:/run/udev:ro
   -v /dev/tty:/dev/i2c-0 -v /dev/tty:/dev/i2c-1 -v /dev/tty:/dev/i2c-2 umrover/travis
   jarvis build -a .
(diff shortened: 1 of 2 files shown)


CameronTressler (approved):
Interesting that lcm send didn't include rover_msgs as a dep before

mmartin4972:
> Interesting that lcm send didn't include rover_msgs as a dep before

I think lcm_tools/common would previously build before lcm_tools/send, so we never noticed the lacking dependency. The order they built in changed for some reason and the error occurred

.travis.yml
```diff
@@ -2,7 +2,7 @@ sudo: required
 services:
 - docker
 script:
-- docker pull umrover/travis
+- docker pull umrover1/travis
```
cgiger00, line 5: where is this exactly?

cgiger00 (approved):

.travis.yml
mmartin4972: https://hub.docker.com/repository/docker/umrover1/travis

niwhsa9 (approved):

PR #814, opened 2021-12-03
[percep] Parallelized find-clear-path

arschallwig:
Implented find-clear-path algorithm in parallel


niwhsa9 (approved):
Truly excellent work... very nicely done to both of you @arschallwig @Aditya-Singhvi 

I'm sure we'll find some bugs during integration testing, so we'll hammer that out in January.

PR #820, opened 2021-12-09
[teleop] add reverse toggle to RA gui

emilyyuan03:
tested with lcm echo and got expected values
Screen Shot 2022-01-12 at 8 21 28 PM


CameronTressler (approved):

jnnanni (approved):

base_station/gui/src/components/ReverseToggle.vue
```diff
@@ -0,0 +1,109 @@
...
+  methods: {
+    setPart: function(enabled) {
+      this.$parent.publish("/reverse_drive", {
+        'type': 'ReverseDrive',
+        'reverse': enabled
+      })
+    }
+  }
+}
+</script>
```
cgiger00, line 109: would be nice if you could add screenshots of how this looks to your gui PRs

jetson/teleop/src/__main__.py
```diff
@@ -129,6 +130,11 @@ def drive_control_callback(channel, msg):
 
         joystick_math(new_motor, magnitude, theta)
 
+        if(reverse):
```
cgiger00, line 133: why is there a reverse check in kill()?

CameronTressler: I don't think there is? Github may just be hiding it but this is in drive_control_callback

CameronTressler:
Change going to be made first in ish-integration. From there it will be merged into main

PR #832, opened 2022-01-10
[nav] Basic Target Caching

zwgold:
Targets are now stored/cached in event perception data is blocked out or zeroed due to noise, preventing us from trying to search for it again despite already having seen it and on the path to the target.

### DIFF
--- a/jetson/nav/obstacle_avoidance/obstacleAvoidanceStateMachine.cpp
+++ b/jetson/nav/obstacle_avoidance/obstacleAvoidanceStateMachine.cpp
@@ -68,7 +68,13 @@ NavState ObstacleAvoidanceStateMachine::run()
 bool ObstacleAvoidanceStateMachine::isTargetDetected ()
 {
     return ( mRover->roverStatus().currentState() == NavState::SearchTurnAroundObs &&
+<<<<<<< HEAD
              mRover->roverStatus().leftTarget().distance >= 0 );
+=======
+             ( mRover->roverStatus().leftTarget().distance >= 0 || 
+                ( mRover->roverStatus().leftTarget().distance < 0 && 
+                    mRover->roverStatus().leftCacheTarget().distance != -1 )));
+>>>>>>> b5fa5c8a (Target Caching System for Nav.)
 }
 
 // The obstacle avoidance factory allows for the creation of obstacle avoidance objects and
--- a/jetson/nav/obstacle_avoidance/obstacleAvoidanceStateMachine.cpp
+++ b/jetson/nav/obstacle_avoidance/obstacleAvoidanceStateMachine.cpp
@@ -68,7 +68,7 @@ NavState ObstacleAvoidanceStateMachine::run()
 bool ObstacleAvoidanceStateMachine::isTargetDetected ()
 {
     // Second check is to see if we have either a valid target, or if we have a valid
-    // cached target
+    // cached target to view
     return ( mRover->roverStatus().currentState() == NavState::SearchTurnAroundObs &&
              ( mRover->roverStatus().leftTarget().distance >= 0 || 
                 ( mRover->roverStatus().leftTarget().distance < 0 && 
--- a/jetson/nav/obstacle_avoidance/obstacleAvoidanceStateMachine.cpp
+++ b/jetson/nav/obstacle_avoidance/obstacleAvoidanceStateMachine.cpp
@@ -70,9 +70,7 @@ bool ObstacleAvoidanceStateMachine::isTargetDetected ()
     // Second check is to see if we have either a valid target, or if we have a valid
     // cached target to view
     return ( mRover->roverStatus().currentState() == NavState::SearchTurnAroundObs &&
-             ( mRover->roverStatus().leftTarget().distance >= 0 || 
-                ( mRover->roverStatus().leftTarget().distance < 0 && 
-                    mRover->roverStatus().leftCacheTarget().distance != -1 ) ) );
+             ( mRover->roverStatus().leftCacheTarget().distance >= 0 ) );
 }
 
 // The obstacle avoidance factory allows for the creation of obstacle avoidance objects and
--- a/jetson/nav/obstacle_avoidance/obstacleAvoidanceStateMachine.cpp
+++ b/jetson/nav/obstacle_avoidance/obstacleAvoidanceStateMachine.cpp
@@ -68,13 +68,9 @@ NavState ObstacleAvoidanceStateMachine::run()
 bool ObstacleAvoidanceStateMachine::isTargetDetected ()
 {
     return ( mRover->roverStatus().currentState() == NavState::SearchTurnAroundObs &&
-<<<<<<< HEAD
-             mRover->roverStatus().leftTarget().distance >= 0 );
-=======
              ( mRover->roverStatus().leftTarget().distance >= 0 || 
                 ( mRover->roverStatus().leftTarget().distance < 0 && 
                     mRover->roverStatus().leftCacheTarget().distance != -1 )));
->>>>>>> b5fa5c8a (Target Caching System for Nav.)
 }
 
 // The obstacle avoidance factory allows for the creation of obstacle avoidance objects and
--- a/jetson/nav/obstacle_avoidance/obstacleAvoidanceStateMachine.cpp
+++ b/jetson/nav/obstacle_avoidance/obstacleAvoidanceStateMachine.cpp
@@ -67,10 +67,12 @@ NavState ObstacleAvoidanceStateMachine::run()
 // Checks that both rover is in search state and that target is detected
 bool ObstacleAvoidanceStateMachine::isTargetDetected ()
 {
+    // Second check is to see if we have either a valid target, or if we have a valid
+    // cached target
     return ( mRover->roverStatus().currentState() == NavState::SearchTurnAroundObs &&
              ( mRover->roverStatus().leftTarget().distance >= 0 || 
                 ( mRover->roverStatus().leftTarget().distance < 0 && 
-                    mRover->roverStatus().leftCacheTarget().distance != -1 )));
+                    mRover->roverStatus().leftCacheTarget().distance != -1 ) ) );
 }
 
 // The obstacle avoidance factory allows for the creation of obstacle avoidance objects and
--- a/jetson/nav/obstacle_avoidance/obstacleAvoidanceStateMachine.cpp
+++ b/jetson/nav/obstacle_avoidance/obstacleAvoidanceStateMachine.cpp
@@ -68,7 +68,7 @@ NavState ObstacleAvoidanceStateMachine::run()
 bool ObstacleAvoidanceStateMachine::isTargetDetected ()
 {
     return ( mRover->roverStatus().currentState() == NavState::SearchTurnAroundObs &&
-             mRover->roverStatus().target().distance >= 0 );
+             mRover->roverStatus().leftTarget().distance >= 0 );
 }
 
 // The obstacle avoidance factory allows for the creation of obstacle avoidance objects and
--- a/jetson/nav/rover.cpp
+++ b/jetson/nav/rover.cpp
@@ -59,7 +59,21 @@ Target& Rover::RoverStatus::rightTarget()
 {
     return mTarget2;
 } // rightTarget()
+<<<<<<< HEAD
 
+=======
+
+Target& Rover::RoverStatus::leftCacheTarget()
+{
+    return mCTarget1;
+} // leftCacheTarget()
+
+Target& Rover::RoverStatus::rightCacheTarget() 
+{
+    return mCTarget2;
+} // rightCacheTarget()
+
+>>>>>>> b5fa5c8a (Target Caching System for Nav.)
 RadioSignalStrength& Rover::RoverStatus::radio() 
 {
     return mSignal;
@@ -69,6 +83,11 @@ unsigned Rover::RoverStatus::getPathTargets()
 {
   return mPathTargets;
 } // getPathTargets()
+
+int& Rover::RoverStatus::getMisses()
+{
+    return countMisses;
+}
 
 // Assignment operator for the rover status object. Does a "deep" copy
 // where necessary.
@@ -90,11 +109,18 @@ Rover::RoverStatus& Rover::RoverStatus::operator=( Rover::RoverStatus& newRoverS
             ++mPathTargets;
         }
     }
+    countMisses = newRoverStatus.getMisses();
     mObstacle = newRoverStatus.obstacle();
     mOdometry = newRoverStatus.odometry();
     mTarget1 = newRoverStatus.leftTarget();
     mTarget2 = newRoverStatus.rightTarget();
+<<<<<<< HEAD
+=======
+    mCTarget1 = newRoverStatus.leftCacheTarget();
+    mCTarget2 = newRoverStatus.rightCacheTarget();
+>>>>>>> b5fa5c8a (Target Caching System for Nav.)
     mSignal = newRoverStatus.radio();
+    countMisses = newRoverStatus.getMisses();
     return *this;
 } // operator=
 
@@ -242,6 +268,50 @@ bool Rover::updateRover( RoverStatus newRoverStatus )
             mRoverStatus.odometry() = newRoverStatus.odometry();
             mRoverStatus.leftTarget() = newRoverStatus.leftTarget();
             mRoverStatus.rightTarget() = newRoverStatus.rightTarget();
+<<<<<<< HEAD
+=======
+
+            // Perform basic caching, include special case when
+            // we store for the first time, but store an empty one
+            // so we don't just have "junk" in the target Cache.
+            if ( mRoverStatus.leftTarget().distance != -1 ) 
+            {
+                mRoverStatus.leftCacheTarget() = mRoverStatus.leftTarget();
+                mRoverStatus.getMisses() = 0;
+            }
+            else 
+            { // if this is empty, then so is rightTarget. increment misses
+                mRoverStatus.getMisses()++;
+            }
+
+            // Note: The only time this updates is ONLY if the leftTarget updates (by default
+            // the first target seen will be stored in leftTarget, and even if only 1 is seen
+            // this won't update, so we should be fine)
+            if ( mRoverStatus.rightTarget().distance != -1 ) 
+            {
+                mRoverStatus.rightCacheTarget() = mRoverStatus.rightTarget();
+                mRoverStatus.getMisses() = 0;
+            }
+
+
+            // If we have too many misses in a row, then we know we definitely are missing the Tag,
+            // so we can set the cache to the left/right target (which both should be empty)
+            // Must also do the reset once we finish a gate search
+            if ( mRoverStatus.getMisses() > mRoverConfig[ "navThresholds" ][ "cacheMissMax" ].GetDouble() )
+            {
+                mRoverStatus.getMisses() = 0;
+                // these will be zero, so set them as such since we know
+                // they definitely aren't there
+                mRoverStatus.leftCacheTarget() = mRoverStatus.leftTarget();
+                mRoverStatus.rightCacheTarget() = mRoverStatus.rightTarget();
+            }
+
+
+            cout << "LEFT CACHE (distance, bearing): (" << mRoverStatus.leftCacheTarget().distance << "," << mRoverStatus.leftCacheTarget().bearing << ")" << endl;
+            cout << "RIGHT CACHE (distance, bearing): (" << mRoverStatus.rightCacheTarget().distance << "," << mRoverStatus.rightCacheTarget().bearing << ")" << endl;
+            cout << "MISSES COUNT: " << mRoverStatus.getMisses() << endl;
+
+>>>>>>> b5fa5c8a (Target Caching System for Nav.)
             mRoverStatus.radio() = newRoverStatus.radio();
             updateRepeater( mRoverStatus.radio() );
             return true;
--- a/jetson/nav/rover.cpp
+++ b/jetson/nav/rover.cpp
@@ -59,9 +59,6 @@ Target& Rover::RoverStatus::rightTarget()
 {
     return mTarget2;
 } // rightTarget()
-<<<<<<< HEAD
-
-=======
 
 Target& Rover::RoverStatus::leftCacheTarget()
 {
@@ -73,7 +70,6 @@ Target& Rover::RoverStatus::rightCacheTarget()
     return mCTarget2;
 } // rightCacheTarget()
 
->>>>>>> b5fa5c8a (Target Caching System for Nav.)
 RadioSignalStrength& Rover::RoverStatus::radio() 
 {
     return mSignal;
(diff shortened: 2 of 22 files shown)


jetson/nav/rover.cpp
```diff
@@ -242,6 +261,47 @@ bool Rover::updateRover( RoverStatus newRoverStatus )
...
+
+            // Note: The only time this updates is ONLY if the leftTarget updates (by default
+            // the first target seen will be stored in leftTarget, and even if only 1 is seen
+            // this won't update, so we should be fine)
+            if ( mRoverStatus.rightTarget().distance != -1 ) 
+            {
+                mRoverStatus.rightCacheTarget() = mRoverStatus.rightTarget();
+                mRoverStatus.getMisses() = 0;
+            }
+
```
niwhsa9, line 286: you don't need to increment misses here in an else statement?

jetson/nav/rover.cpp
```diff
@@ -242,6 +261,47 @@ bool Rover::updateRover( RoverStatus newRoverStatus )
...
+
+            // If we have too many misses in a row, then we know we definitely are missing the Tag,
+            // so we can set the cache to the left/right target (which both should be empty)
+            // Must also do the reset once we finish a gate search
+            if ( mRoverStatus.getMisses() > mRoverConfig[ "navThresholds" ][ "cacheMissMax" ].GetDouble() )
+            {
+                mRoverStatus.getMisses() = 0;
+                // these will be zero, so set them as such since we know
```
niwhsa9, line 294: zero or negative one?

jetson/nav/rover.cpp
```diff
@@ -242,6 +261,47 @@ bool Rover::updateRover( RoverStatus newRoverStatus )
...
+                // they definitely aren't there
+                mRoverStatus.leftCacheTarget() = mRoverStatus.leftTarget();
+                mRoverStatus.rightCacheTarget() = mRoverStatus.rightTarget();
```
niwhsa9, line 297: i think it would be better to explicitly assign these to the value of the missing target.

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -116,6 +122,12 @@ NavState SearchStateMachine::executeRoverWait()
                                        mRover->roverStatus().odometry().bearing_deg );
         return NavState::TurnToTarget;
     }
+    else if( mRover->roverStatus().leftTarget().distance == -1 &&
```
niwhsa9, line 125: copy-pasta error? should this say rightTarget?

jetson/nav/obstacle_avoidance/obstacleAvoidanceStateMachine.cpp
```diff
@@ -68,7 +68,9 @@ NavState ObstacleAvoidanceStateMachine::run()
 bool ObstacleAvoidanceStateMachine::isTargetDetected ()
 {
     return ( mRover->roverStatus().currentState() == NavState::SearchTurnAroundObs &&
```
mmartin4972, line 70: Are you checking that you are in a turn around obstacle state here because you're assuming that you'll only be able to see the target if it is being detected as an obstacle?

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -182,6 +200,11 @@ NavState SearchStateMachine::executeSearchDrive()
                                            mRover->roverStatus().odometry().bearing_deg );
         return NavState::TurnToTarget;
     }
+    else if( mRover->roverStatus().leftTarget().distance == -1 && mRover->roverStatus().leftCacheTarget().distance >= 0 ) {
```
niwhsa9, line 203: same as above, why is this duplicated?

jetson/nav/obstacle_avoidance/obstacleAvoidanceStateMachine.cpp
```diff
@@ -68,7 +68,9 @@ NavState ObstacleAvoidanceStateMachine::run()
...
-             mRover->roverStatus().leftTarget().distance >= 0 );
+             ( mRover->roverStatus().leftTarget().distance >= 0 || 
+                ( mRover->roverStatus().leftTarget().distance < 0 &&
```
mmartin4972, line 72: Is this second check for when you see the tag, but don't get an accurate distance? Could you add a comment explaining that if so?

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -213,19 +236,40 @@ NavState SearchStateMachine::executeTurnToTarget()
...
+            {
+                return NavState::DriveToTarget;
+            }
+            updateTargetDetectionElements( mRover->roverStatus().leftCacheTarget().bearing,
+                                       mRover->roverStatus().odometry().bearing_deg );
+            return NavState::TurnToTarget;
+        }
+        else {
+            cerr << "Lost the target. Continuing to turn to last known angle\n";
+            if( mRover->turn( mTargetAngle + mTurnToTargetRoverAngle ) )
```
niwhsa9, line 252: is the last known angle being updated when using a cached target?

niwhsa9 (changes requested):

jetson/nav/rover.cpp
```diff
@@ -242,6 +261,47 @@ bool Rover::updateRover( RoverStatus newRoverStatus )
...
+            }
+            else 
+            { // if this is empty, then so is rightTarget. increment misses
+                mRoverStatus.getMisses()++;
+            }
+
+            // Note: The only time this updates is ONLY if the leftTarget updates (by default
+            // the first target seen will be stored in leftTarget, and even if only 1 is seen
+            // this won't update, so we should be fine)
+            if ( mRoverStatus.rightTarget().distance != -1 )
```
mmartin4972, line 281: Move this if check to inside the check on line 268, so we know that this will only update if the left one updates.

jetson/nav/rover.cpp
```diff
@@ -242,6 +261,47 @@ bool Rover::updateRover( RoverStatus newRoverStatus )
...
+            {
+                mRoverStatus.getMisses() = 0;
+                // these will be zero, so set them as such since we know
+                // they definitely aren't there
+                mRoverStatus.leftCacheTarget() = mRoverStatus.leftTarget();
+                mRoverStatus.rightCacheTarget() = mRoverStatus.rightTarget();
+            }
+
+
+            cout << "LEFT CACHE (distance, bearing): (" << mRoverStatus.leftCacheTarget().distance << "," << mRoverStatus.leftCacheTarget().bearing << ")" << endl;
```
mmartin4972, line 301: Are these cout statements just for testing?

jetson/nav/rover.cpp
zwgold: I only have it increment on leftTarget misses, since the was the original detection was designed (and to keep the "queue" style of storing targets) was the first target detected is always going to be stored as the left one until another target is detected. Since if the leftTarget misses, that means so will the rightTarget. You bring up a point, if we see the leftTarget but don't see the rightTarget, that could be an issue.

jetson/nav/rover.cpp
zwgold: Should be zero after testing

jetson/nav/search/searchStateMachine.cpp
zwgold: No, this is fine as is

zwgold: No, at this point, the rover is trying to wait before locating the first target (which is the `leftTarget` one)

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -81,6 +81,12 @@ NavState SearchStateMachine::executeSearchSpin()
                                            mRover->roverStatus().odometry().bearing_deg );
```
mmartin4972, line 81: Put lines 78-89 in a function and then use that function throughout this file

jetson/nav/search/searchStateMachine.cpp
zwgold: Yes, in the above lines for when we can't find the current target (function `updateTargetDetectionElements` on line 246 updates the turn so that in the event we lose it, it will turn to the cached one instead)

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -259,18 +305,56 @@ NavState SearchStateMachine::executeDriveToTarget()
...
     if( mRover->roverStatus().rightTarget().distance > 0 )
     {
+        // Valid leftTarget, perform comp.
         if( mRover->roverStatus().leftTarget().distance > mRover->roverStatus().rightTarget().distance ) 
         {
             distance = mRover->roverStatus().rightTarget().distance;
             bearing = mRover->roverStatus().rightTarget().bearing + mRover->roverStatus().odometry().bearing_deg;
         }
+        // If invalid leftTarget, check Cached target
+        else if( mRover->roverStatus().leftTarget().distance < 0 && mRover->roverStatus().leftCacheTarget().distance != -1 )
```
mmartin4972, line 329: Isn't it impossible to have an invalid left target if you have a valid right target? How are the left target and right target determined?

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -259,18 +305,56 @@ NavState SearchStateMachine::executeDriveToTarget()
...
             distance = mRover->roverStatus().rightTarget().distance;
             bearing = mRover->roverStatus().rightTarget().bearing + mRover->roverStatus().odometry().bearing_deg;
         }
+        // If invalid leftTarget, check Cached target
+        else if( mRover->roverStatus().leftTarget().distance < 0 && mRover->roverStatus().leftCacheTarget().distance != -1 ) 
+        {
+            if( mRover->roverStatus().leftCacheTarget().distance > mRover->roverStatus().rightTarget().distance ) 
+            {
+                cout << "USING CACHE" << endl;
+                distance = mRover->roverStatus().rightTarget().distance;
```
mmartin4972, line 334: You aren't using the cached info here. Did you intend to?

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -259,18 +305,56 @@ NavState SearchStateMachine::executeDriveToTarget()
...
+        bearing = mRover->roverStatus().leftCacheTarget().bearing + mRover->roverStatus().odometry().bearing_deg;
+        cout << "USING CACHE" << endl;
+    }
+
     // Executes the logic for driving with 0, 1, or 2 targets in sight
     // If we have a second target detected, determine which post is closer
     // If the distance to the second target is less than the first,
     // set our variables to the target 2's distance and bearing
     // Else, use the initialized values from target 1 when driving
     if( mRover->roverStatus().rightTarget().distance > 0 )
```
mmartin4972, line 320: Shouldn't this if and else if statement be inside of the if statement on line 309, if you will only have a valid right tag if there is a valid left tag?

jetson/nav/rover.cpp
niwhsa9: meant more from a readability perspective. But if theres a compelling reason to keep it this way that I haven't thought of thats fine too

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -259,18 +305,56 @@ NavState SearchStateMachine::executeDriveToTarget()
...
+            if( mRover->roverStatus().leftCacheTarget().distance > mRover->roverStatus().rightTarget().distance ) 
+            {
+                cout << "USING CACHE" << endl;
+                distance = mRover->roverStatus().rightTarget().distance;
+                bearing = mRover->roverStatus().rightTarget().bearing + mRover->roverStatus().odometry().bearing_deg;
+            }
+        }
+    }
+    // Consider Cached rightTarget (in event we lose it for a sec)
+    else if ( mRover->roverStatus().rightTarget().distance < 0 && mRover->roverStatus().rightCacheTarget().distance != -1 )
```
mmartin4972, line 340: The logic in this else if is really hard to follow. Can you try to make it more clear?

mmartin4972 (changes requested):

jetson/nav/rover.cpp
zwgold: My bad, I'll make sure to remove them (wanted to keep them in for when we would test in-person, but I understand that it's better to remove them for style)

jetson/nav/rover.cpp
zwgold: Honestly fair point, I'll just set them equal to the default struct for an empty target.

jetson/nav/obstacle_avoidance/obstacleAvoidanceStateMachine.cpp
zwgold: No, this is a state for checking that if we happen to already be turning around an obstacle, but we also happen to see the target in the view while turning (this would make it so that instead of us going around the obstacle and resuming a search, if we are able to locate it while turning around the obstacle, we can just turn to that target and drive to it)

jetson/nav/search/searchStateMachine.cpp
zwgold: Good catch, I wanted to catch all possible cases, but forgot that if leftTarget is null, rightTarget is guaranteed to be null.

jetson/nav/search/searchStateMachine.cpp
zwgold: This was just a comment, I'll remove it in the next pull

jetson/nav/search/searchStateMachine.cpp
zwgold: For the other part though, we still need to check if leftTarget is valid, since rightTarget being invalid does not imply that leftTarget is, if that makes sense?

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -354,6 +417,26 @@ void SearchStateMachine::insertIntermediatePoints()
     }
 } // insertIntermediatePoints()
 
+bool SearchStateMachine::checkTurnStatus() {
+    double updateBearing = ( mRover->roverStatus().leftTarget().distance == -1
```
cgiger00, line 421: here again is there a reason not to just use the cached target's bearing?

jetson/nav/obstacle_avoidance/obstacleAvoidanceStateMachine.cpp
mmartin4972: That makes sense. Thanks

mmartin4972 (commented):
Am going to wait for Carmen and Ashwin to respond to their reviews

jetson/nav/search/searchStateMachine.cpp
niwhsa9: great, thx

jetson/nav/rover.cpp
niwhsa9: it should treat each target's cache as independent, if one is available and the other isn't you should use the cache for the one that isn't and the ground truth for the one that is.

jetson/nav/obstacle_avoidance/obstacleAvoidanceStateMachine.cpp
```diff
@@ -67,8 +67,12 @@ NavState ObstacleAvoidanceStateMachine::run()
 // Checks that both rover is in search state and that target is detected
 bool ObstacleAvoidanceStateMachine::isTargetDetected ()
 {
+    // Second check is to see if we have either a valid target, or if we have a valid
+    // cached target
     return ( mRover->roverStatus().currentState() == NavState::SearchTurnAroundObs &&
-             mRover->roverStatus().leftTarget().distance >= 0 );
+             ( mRover->roverStatus().leftTarget().distance >= 0 || 
+                ( mRover->roverStatus().leftTarget().distance < 0 && 
+                    mRover->roverStatus().leftCacheTarget().distance != -1 ) ) );
```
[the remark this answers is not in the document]
niwhsa9: better yet define a `const int NO_TARGET = -1` and any other constants you need for things like this.

jetson/nav/rover.cpp
```diff
@@ -60,6 +62,16 @@ Target& Rover::RoverStatus::rightTarget()
     return mTarget2;
 } // rightTarget()
 
+Target& Rover::RoverStatus::leftCacheTarget()
+{
+    return mCTarget1;
```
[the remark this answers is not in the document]
zwgold: Had named it due to original style (originally called target and target2) but didn't rename the private variables. Will fix

jetson/nav/rover.cpp
```diff
@@ -70,6 +82,11 @@ unsigned Rover::RoverStatus::getPathTargets()
   return mPathTargets;
 } // getPathTargets()
 
+int& Rover::RoverStatus::getMisses()
```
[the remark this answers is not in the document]
zwgold: This was the design pattern Andrew had done for past implementations, since we want to return the current count of misses, but also be able to modify it without having 2 separate functions, but rather 1 that essentially does both,

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -213,19 +209,41 @@ NavState SearchStateMachine::executeTurnToTarget()
...
+                return NavState::TurnedToTargetWait;
+            }
+            return NavState::TurnToTarget;
         }
-        return NavState::TurnToTarget;
     }
-    if( mRover->turn( mRover->roverStatus().leftTarget().bearing +
+
+    // Consider if we have a valid cache. If we do, use it. Otherwise, 
+    // can just use the regular leftTarget since it will be filled or empty
```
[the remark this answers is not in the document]
zwgold: We need to flush it out at somepoint, so I felt after sometime we need to reset the cache. This might be a rare case, but the event that we have a false read of a tag, we don't want to permanently store it in the cache, then use that cache always (resulting in it trying to find a non-existent tag).

In any case, I did it just to flush out info, mostly to avoid storing an old cache (from say, an earlier Tag identification)

jetson/nav/obstacle_avoidance/obstacleAvoidanceStateMachine.cpp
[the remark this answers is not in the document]
zwgold: I can define a variable in our `config.json` for this, good call

jetson/nav/rover.cpp
```diff
@@ -242,14 +263,37 @@ bool Rover::updateRover( RoverStatus newRoverStatus )
             mRoverStatus.odometry() = newRoverStatus.odometry();
             mRoverStatus.leftTarget() = newRoverStatus.leftTarget();
             mRoverStatus.rightTarget() = newRoverStatus.rightTarget();
+
+            // Cache Target if we had detected one
+            if( mRoverStatus.leftTarget().distance != -1 ) 
+            {
+                mRoverStatus.leftCacheTarget() = mRoverStatus.leftTarget();
+                if( mRoverStatus.rightTarget().distance != -1 )
```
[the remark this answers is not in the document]
zwgold: The rightTarget is only filled if the leftTarget is filled (by default, if only one seen, it goes into leftTarget)

jetson/nav/rover.cpp
zwgold: Remember that most of the time, we don't need to worry about the rightTarget, since that is only used for the gate (which there's only 1 of, as opposed to 4 in the past); and that mostly we;ll be needing to worry about Tagged posts. I'll consider changing it in the future since you bring up a good point

jetson/nav/search/searchStateMachine.cpp
zwgold: Cleaned that up, added some comments. Let me know if more is needed to clarify

jetson/nav/obstacle_avoidance/obstacleAvoidanceStateMachine.cpp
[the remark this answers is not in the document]
cgiger00: sure variables would be ideal

jetson/nav/rover.cpp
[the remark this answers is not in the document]
cgiger00: eh

jetson/nav/search/searchStateMachine.cpp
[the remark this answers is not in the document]
cgiger00: aren't false positives like one in life time rare events? even so 20 frames is fairly quick to clear that. i think it'll clean up a lot of this by just using the cache & only setting it to left target (or right) if it sees smth else increase misses, & if the misses are above the threshold reset it, & then everywhere else only use the cache.

jetson/nav/obstacle_avoidance/obstacleAvoidanceStateMachine.cpp
cgiger00: ^

jetson/nav/utilities.cpp
```diff
@@ -129,11 +129,12 @@ void clear( deque<Waypoint>& aDeque )
 bool isTargetReachable( Rover* rover, const rapidjson::Document& roverConfig )
 {
     double distToTarget = rover->roverStatus().leftTarget().distance;
-    if ( rover->roverStatus().leftTarget().distance < 0 && rover->roverStatus().leftCacheTarget().distance != -1 )
+    if ( rover->roverStatus().leftTarget().distance < 0 && rover->roverStatus().leftCacheTarget().distance != 
+        roverConfig[ "navThresholds" ][ "targetDistance" ].GetDouble() )
```
cgiger00, line 133: isnt this supposed to be noTargetDistance?

cgiger00 (changes requested):

jetson/nav/rover.cpp
```diff
@@ -11,6 +11,8 @@ Rover::RoverStatus::RoverStatus()
     : mCurrentState( NavState::Off )
 {
     mAutonState.is_auton = false;
+    mCTarget1 = {-1, 0, 0};
+    mCTarget2 = {-1, 0, 0};
```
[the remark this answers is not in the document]
zwgold: This should be specified in the new one (now renamed with left and right)

jetson/nav/search/searchStateMachine.cpp
zwgold: Thought about it and honestly you are right, caches will be up to date regardless, so let's just replace all instances of leftTarget/rightTarget with their cached ones; I just included them to be sure of any inconsistencies (worst case if this fails, we add them back even if it's ugly)

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -256,20 +263,20 @@ NavState SearchStateMachine::executeDriveToTarget()
...
     // Executes the logic for driving with 0, 1, or 2 targets in sight
     // If we have a second target detected, determine which post is closer
     // If the distance to the second target is less than the first,
     // set our variables to the target 2's distance and bearing
     // Else, use the initialized values from target 1 when driving
-    if( mRover->roverStatus().rightTarget().distance > 0 )
+    if( mRover->roverStatus().rightCacheTarget().distance != mRoverConfig[ "navThresholds" ][ "noTargetDist" ].GetDouble() )
     {
-        if( mRover->roverStatus().leftTarget().distance > mRover->roverStatus().rightTarget().distance ) 
+        if( mRover->roverStatus().leftCacheTarget().distance > mRover->roverStatus().rightCacheTarget().distance )
```
mmartin4972, line 276: Combine this with the if statement on line 274

mmartin4972 (changes requested):

jetson/nav/search/searchStateMachine.cpp
mmartin4972: Looks great. Thanks

mmartin4972 (approved):

niwhsa9 (approved):

zwgold:
Finished running on test suite with and without caching, things are working fine, no differences in performance 👍

cgiger00 (approved):

PR #838, opened 2022-01-12
fixed scope error

niwhsa9:
The sl::Mat which owns the PC data retrieved from the ZED was going out of scope before the detector could run causing it to be destructed and invalidating the pointer in the GPU_Cloud.


rbridges12 (approved):

arschallwig (approved):
Looks good

PR #841, opened 2022-01-14
GPS Spoof for localization filter

rbridges12:
Added a new filter mode that passes bearing data from the IMU and a static GPS position to odometry. We can use this for integration testing near the AR tag without going outside or having the GPS connected.

### DIFF
--- a/jetson/filter/src/__main__.py
+++ b/jetson/filter/src/__main__.py
@@ -149,6 +149,10 @@ class SensorFusion:
         new_imu = IMUData.decode(msg)
         self.imu.update(new_imu)
 
+        # In BearingOnlyPipe mode, init filter once IMU data is ready
+        if self.filter is None and self.imu.ready() and self.config["FilterType"] == "BearingOnlyPipe":
+            self.filter = "BearingOnlyPipe"
+
     def _navStatusCallback(self, channel, msg):
         new_nav_status = NavStatus.decode(msg)
         self.nav_state = new_nav_status.nav_state_name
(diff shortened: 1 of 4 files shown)


jetson/mp9250_bridge/main.cpp
```diff
@@ -251,7 +251,7 @@ int main(void)
     char r_buf[1024];
     bzero(r_buf,1024);
 
-    fd = uart_open(fd,"/dev/ttyS1");/*串口号/dev/ttySn,USB口号/dev/ttyUSBn */ 
+    fd = uart_open(fd,"/dev/ttyUSB0");/*串口号/dev/ttySn,USB口号/dev/ttyUSBn */
```
niwhsa9, line 254: Since we are using the FTDI UART->USB for everything I think this is appropriate.

niwhsa9 (approved):

jetson/filter/src/__main__.py
```diff
@@ -356,6 +363,24 @@ async def run(self):
...
                                                         ref_lat=self.config["RefCoords"]["lat"],
                                                         ref_long=self.config["RefCoords"]["long"])
+
+                # Simulates a fixed position at the TargetCoords specified in config,
+                # only passes through valid bearing data
+                elif self.config["FilterType"] == "BearingOnlyPipe":
+                    bearing = self._getFreshBearing()
+                    if bearing is None:
+                        continue
+                    decimal_pos = self.config["TargetCoords"]
```
mmartin4972, line 373: Are we able to change retrieving the TargetCoords properly v. retrieving the TargetCoords from the config file, by changing a build flag or the config file?

mmartin4972 (changes requested):
Let's not merge this into auton-integration yet. Let's create a new branch in the mrover-workspace, so we can test it on this separate branch first

cgiger00 (approved):

mmartin4972 (approved):

PR #846, opened 2022-01-16
Integration fix obs init

mmartin4972:
Initialized percep obstacle LCM


niwhsa9 (commented):
is this supposed to go into main or auton-integration?

PR #849, opened 2022-01-16
Added New Nav State indication in lower left of gui, removed camera box

jnnanni:


base_station/gui/src/components/AutonTask.vue
```diff
@@ -179,6 +180,32 @@ export default {
   },
 
   created: function () {
+    setInterval(() => {
+      if(this.nav_status.nav_state_name == "Off"){
+        this.nav_state_color = "#4695FF"
```
cgiger00, line 185: is this blue? can this be a variable?

cgiger00 (changes requested):

CameronTressler:
Similar changes merged in a different PR

PR #858, opened 2022-01-21
[percep] Unit testing

chrsdavis:
Function overview:

constructor -> test_input_file -> test -> test_print

Utility functions: test_awake() and calculate_intersection()

If ObsDetector has mode TEST, then it calls test_input_file() in the constructor. test_input_file() reads in a user-input file path to a .pcd file. It generates a "ground truth" and saves the unprocessed clouds, which it passes to test(). Test runs each obstacle return and cloud it receives (a vector). It passes the clouds through the pipeline and compares it with the "ground truth."

the test_awake() means the viewer is displayed by freezing the thread until the function returns. Currently, it's just a set amount of seconds for this to happen.

Calculate intersection just calculates an intersection.

Simplescenegenerator is supposed to generate hand-made ground truth pcd's (and obs returns), however the I/O is so slow that the resolution has to be magnitudes smaller than real data. This means that RANSAC and passZ both filter out the truths as noise. Also, these are usually simply not detected by the main pipeline. This is what future work needs to go over.


cgiger00:
does not compile, member has left team, other things are more critical atm

PR #868, opened 2022-01-24
serial parser for imu

cgiger00:


niwhsa9:
should we rebase though, seems like there is some extra stuff in here?

cgiger00:
yeah i rebased with main...will redo it with auton-integration, we should probably rebase that branch with main sometime soon tho as well

rbridges12 (approved):

PR #874, opened 2022-01-26
Arduino code for sending IMU data

rbridges12:
Arduino program to read data from the ICM-20948 IMU and send it over serial. Also slightly changed the IMU driver to work with this.


niwhsa9 (approved):
lgtm

cgiger00 (approved):

PR #876, opened 2022-01-27
Obs detect update

mmartin4972:
* Created new camera interface to break apart camera logic
* Added config file
* Added AR Tag processing
* Moved data sets to the perception-files [repository](https://github.com/umrover/perception-files)
---
* TODO:
* The only supported resolution right now is 320x180 because of how we are opening the camera
* Get Jarvis to compile with cmake
* Get LCMs working
* Figure out how to get the config file to work when located in the config folder


mmartin4972:
Configuration on this branch is not correct.

PR #885, opened 2022-01-29
Let user specify recording directory in viewer

RyanBoring:


niwhsa9:
@RyanBoring can u rebase with gpuObsDetection head?

PR #887, opened 2022-01-29
Add Raw+Bounding+Bearing GUI layer

RyanBoring:


niwhsa9:
@RyanBoring can u rebase with gpuObsDetection head?

PR #888, opened 2022-01-30
[sim] Percep noise + GPS noise simulator with threshold adjustment

danieljiang520:


zwgold:
image
Is there any way to fix this so that we have the decimals truncated? (i.e. increments by 0.1?)

cgiger00 (approved):
looks good to me

zwgold (dismissed):

mmartin4972 (changes requested):
Why are you merging this into auton-integration? Do you need to integration test it? Can't it just go right into main if it is working good?

zwgold (approved):

mmartin4972 (changes requested):
There were a few comments that seemed left over from testing. Can you remove those or was there a reason for keeping them in? There were some eslint and console.log comments that I think you can remove.

mmartin4972 (approved):
Daniel told me the following regarding the eslint comments, so we are going to leave them in. 

> I left the eslint comments in constants.ts because that’s the only way I found to let eslint ignore the “magic numbers” in the constant array

zwgold (approved):
@cgiger00 seems solid, can probs merge at this point

cgiger00:
is out of date w main...had to remerge it :/

cgiger00 (approved):
please include screenshots of the sim with it's changes next time!

PR #891, opened 2022-02-01
[esw] Rolling rover drive code 

cgiger00:
resolves # (none)

Drive code for rolling rover, updated to run with downgraded firmware version compatible with 3 odrives


tabiosg (approved):

jetson/odrive_bridge/src/__main__.py
```diff
@@ -515,28 +393,15 @@ def _requested_state(self, state):
...
-        global legal_controller
         if (axis == "LEFT"):
-            # TEMPORARY FIX FOR ROLLING ROVER SINCE
-            # middle left odrive IS 2x more than the rest bc of the 48V maxon
-            # TODO - fix when this is no longer the case!
-            if (legal_controller == 1):
-                self.front_axis.controller.input_vel = vel * 100
-            else:
-                self.front_axis.controller.input_vel = vel * 50
+            self.front_axis.controller.input_vel = -vel * 3
```
CameronTressler, line 397: Out of curiosity, what is this code doing, specifically with the 3?

cgiger00: lol that looks like smth wrong happened during the rebase, good catch

CameronTressler (dismissed):

CameronTressler (approved):

tabiosg (approved):

jetson/odrive_bridge/src/__main__.py
```diff
@@ -233,15 +224,17 @@ def connect(self):
         # odrive 1 --> middle motors
         # odrive 2 --> back motors
 
-        odrives = ["335D36623539", "335B36563539", "335536553539"]
+        odrives = ["206E37635753", "335B36563539", "335536553539"]
```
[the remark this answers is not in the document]
cgiger00: good point

tabiosg (dismissed):

jetson/odrive_bridge/src/__main__.py
```diff
@@ -234,14 +225,16 @@ def connect(self):
         # odrive 2 --> back motors
 
         odrives = ["335D36623539", "335B36563539", "335536553539"]
+
         id = odrives[legal_controller]
 
         print(id)
+        # import pdb; pdb.set_trace()
```
mmartin4972, line 232: Why was this added?

jetson/odrive_bridge/src/__main__.py
```diff
@@ -83,18 +90,14 @@ def lcmThreaderMan():
...
-        except NameError:
-            pass
-        except AttributeError:
-            pass
-        except Exception:
+        except (NameError, AttributeError, fibre.protocol.ChannelBrokenException):
             pass
 
 
 events = ["disconnected odrive", "disarm cmd", "arm cmd", "odrive error"]
```
mmartin4972, line 97: You should make enums for the events and the states

jetson/odrive_bridge/src/__main__.py
```diff
@@ -50,23 +50,30 @@ def main():
     # starting state is DisconnectedState()
     # start up sequence is called, disconnected-->disarm-->arm
 
+    prev_comms = False
+
     while True:
         watchdog = t.clock() - start_time
         if (watchdog > 1.0):
```
mmartin4972, line 57: Did you lint this? I think Pylint complains about having parentheses in conditional statements

jetson/odrive_bridge/src/__main__.py
```diff
@@ -258,28 +251,25 @@ def on_event(self, event):
...
-        try:
-            errors = modrive.check_errors()
-            modrive.watchdog_feed()
-        except Exception:
-            errors = 0
-            lock.acquire()
-            self.on_event("disconnected odrive")
-            lock.release()
-            print("unable to check errors of unplugged odrive")
+        if (str(self.state) == "ArmedState"):
```
mmartin4972, line 254: Using enums would clean up comparisons like this

jetson/odrive_bridge/src/__main__.py
```diff
@@ -234,14 +225,16 @@ def connect(self):
...
         id = odrives[legal_controller]
 
         print(id)
+        # import pdb; pdb.set_trace()
         odrive = odv.find_any(serial_number=id)
 
         print("found odrive")
         modrive = Modrive(odrive)  # arguments = odr
-        modrive.set_current_lim(100)
+        modrive.set_current_lim(4)
```
mmartin4972, line 237: Is this related to the CURRENT_LIM constant in Modrive? Could you use constant for these values instead of just setting them?

jetson/odrive_bridge/src/__main__.py
```diff
@@ -50,23 +50,30 @@ def main():
     # starting state is DisconnectedState()
     # start up sequence is called, disconnected-->disarm-->arm
 
+    prev_comms = False
+
     while True:
         watchdog = t.clock() - start_time
```
mmartin4972, line 56: Can you add a comment explaining what the watchdog variable is? It's a bit difficult to understand what it is being used for

jetson/odrive_bridge/src/__main__.py
```diff
@@ -258,28 +251,25 @@ def on_event(self, event):
...
-        except Exception:
-            errors = 0
-            lock.acquire()
-            self.on_event("disconnected odrive")
-            lock.release()
-            print("unable to check errors of unplugged odrive")
+        if (str(self.state) == "ArmedState"):
+            try:
+                errors = modrive.check_errors()
+                modrive.watchdog_feed()
```
mmartin4972, line 257: Don't you need to declare modrive and lock as a global to reference them here? Aren't you just creating it locally rn?

mmartin4972 (changes requested):

jetson/odrive_bridge/src/__main__.py
cgiger00: debugging purposes

jetson/odrive_bridge/src/__main__.py
cgiger00: enums are annoying in python, this is the work around that's about as efficient/readable

jetson/odrive_bridge/src/__main__.py
cgiger00: passes the linter & travis so yes

jetson/odrive_bridge/src/__main__.py
cgiger00: state is a class variable, not exactly sure what you mean

jetson/odrive_bridge/src/__main__.py
cgiger00: resolved

jetson/odrive_bridge/src/__main__.py
cgiger00: no, arguably the only place you need it declared globally is in connect when it's assigned to, if it's read-only in a function python assumes it's part of the global scope. it's probably declared globally everywhere else bc the original authors were being careful if i had to guess. also update is in a lock already

jetson/odrive_bridge/src/__main__.py
mmartin4972: https://docs.python.org/3/library/enum.html 
They don't seem that annoying. It looks like you just create a class and reference it throughout. Why do you say they are annoying?

jetson/odrive_bridge/src/__main__.py
mmartin4972: I was thinking that if you had made the events and state variables enums, then you wouldn't need to be casting your state to a string and could just compare it with an enum.

mmartin4972 (changes requested):
I'm still not convinced that enums are difficult to use in python. Is the documentation I was looking at for a different version than what I was looking at here: https://docs.python.org/3/library/enum.html ?

jetson/odrive_bridge/src/__main__.py
cgiger00: i don't cast it, state is a class and str is like a function/property of the class, so for each state i would need an extra object in the class that's an enum defining it, such as state.NAME that returns that enum, which seems like pretty much the same thing in my opinion. i can change the event names to be enums although that doesn't do anything to the code clarity imo

cgiger00:
see comment above on why i don't think they're useful in this implementation

jetson/odrive_bridge/src/__main__.py
cgiger00: would be annoying to get working in this code more so, i'd be adding a class within a class to call a function that acts the same as .str which regardless is in there

cgiger00: i think could be rewritten to be if self.state == ArmState no casting at all, not sure how python will handle the comparison though would have to test

mmartin4972: Gotcha. I didn't realize that self.state was a class and had functionality beyond just being a variable that you are comparing against

mmartin4972: Seems like you could make a getter for interfacing with the class, but what you have is clear enough

jetson/odrive_bridge/src/__main__.py
mmartin4972: Yeah, I didn't realize it was a class. That's my bad. Please change the event names to enums, since I do think that would be better coding practice

mmartin4972 (changes requested):
I misunderstood your use of the state variable. Thank you for clarifying. Please set the events variables to be enums and the rest looks good

jetson/odrive_bridge/src/__main__.py
cgiger00: sure

jetson/odrive_bridge/src/__main__.py
```diff
@@ -233,15 +232,16 @@ def connect(self):
         # odrive 1 --> middle motors
         # odrive 2 --> back motors
 
-        odrives = ["335D36623539", "335B36563539", "335536553539"]
+        odrives = ["335D36623539", "335B36563539", "2066377F5753"]
```
tabiosg, line 235: update back to regular ids

tabiosg (dismissed):

jetson/odrive_bridge/src/__main__.py
```diff
@@ -443,30 +403,29 @@ def disable_watchdog(self):
             self.back_axis.config.watchdog_timeout = 0
             self.front_axis.config.enable_watchdog = False
             self.back_axis.config.enable_watchdog = False
-        except Exception as e:
-            print("Failed in disable_watchdog. Error:")
-            print(e)
+        except fibre.protocol.ChannelBrokenException:
```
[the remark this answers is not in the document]
cgiger00: it wouldn't pass travis without passing the linter. the linter fails when you do an except (nothing): so except fibre.protocol.ChannelBrokenException is fine

jetson/odrive_bridge/src/__main__.py
tabiosg: oops didn't realize that we are using 2066377F5753 now

tabiosg (approved):

CameronTressler (approved):

cgiger00:
> I misunderstood your use of the state variable. Thank you for clarifying. Please set the events variables to be enums and the rest looks good

fixed

cgiger00 (commented):
enum changes

mmartin4972 (approved):

PR #895, opened 2022-02-04
Merge AR Tag Code

qhdwight:


qhdwight:
@niwhsa9 Should be ready to merge after testing on Zed, please review

qhdwight:
I also formatted code

qhdwight:
Also see the `Dockerfile` in `./tools`, builds with both Jarvis and without it. Has OpenCV and LCM

PR #896, opened 2022-02-05
Arduino code for new BNO055 IMU

rbridges12:
Arduino program to read data from the BNO055 IMU and send it over serial, plus a readme for setting it up.


cgiger00 (approved):
looks good to me

PR #901, opened 2022-02-07
[nav] Fixes from rover testing on 2/6

zwgold:
Code to make sure we are initializing the perception data (targets and obstacles) properly to avoid issues if we are not getting LCM information from percep, and to reduce joystick power by 25%.


cgiger00 (approved):
lgtm

PR #902, opened 2022-02-08
Updated launch script

mmartin4972:
- TODO: test ssh capabilities

### DIFF
--- a/jarvis_files/launch_scripts/loc
+++ b/jarvis_files/launch_scripts/loc
@@ -6,27 +6,26 @@ xdotool windowfocus $WID
 
 # Enter commands you want to be executed in the window here
 
-# Run Localization
+# Check If need to ssh first
+if [[ $1 ]]; then
+    ./jarvis_files/launchScripts/ssh $WID
+fi
 xdotool key type "sudo ./jarvis exec jetson/gps"
 xdotool key KP_Enter
 sleep 0.4s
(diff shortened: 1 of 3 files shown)


jarvis_files/launch_scripts/loc
```diff
@@ -6,27 +6,26 @@ xdotool windowfocus $WID
...
+    ./jarvis_files/launchScripts/ssh $WID
+fi
 xdotool key type "sudo ./jarvis exec jetson/gps"
 xdotool key KP_Enter
 sleep 0.4s
 xdotool key type "mrover"
 xdotool key KP_Enter
 
-#Open a new tab for Obstacle LCM
+#Open a new tab
```
zwgold, line 19: Just small thing: but in the comments let's keep the spacing consistent, and at least add what tab is being run

jarvis_files/launch_scripts/loc
```diff
@@ -6,27 +6,26 @@ xdotool windowfocus $WID
...
 xdotool key ctrl+shift+t
 # Check If need to ssh first
 if [[ $1 ]]; then
```
zwgold, line 22: this is more for my understanding, but if you are always going to run this, is it possible just to have the command run without the need of an if-fi statement?

mmartin4972:
Deprecated

PR #918, opened 2022-02-13
Add functionality for Camera buttons on GUI to send LCM

amszuch:


base_station/gui/src/components/Cameras.vue
```diff
@@ -109,8 +109,9 @@
       return {
         dual_stream: false,
         pi_index_1: -1,
-        pi_index_2: -1
-      }
+        pi_index_2: -1,
```
tabiosg, line 112: Are we able to rename for pi_index to something else that is less confusing for the future (since we no longer use pi cams)?

tabiosg (changes requested):
Mainly just rename from pi to something else so stuff is not confusing for future

tabiosg (approved):

PR #921, opened 2022-02-13
[admin] removed python3.7 from ansible scripts

cgiger00:
Migrating over to ubuntu20.04 soon enough, reverting this to work with 3.6 until that update goes through since it's causing issues


tabiosg (approved):

mmartin4972:
Is the default 3.7 for Ubuntu 20? By having it default to python3 when installing using Ubuntu 20 will it use Python 3.7?

cgiger00:
3.8

mmartin4972 (approved):
Sounds good

PR #922, opened 2022-02-15
Changed obstacle avoidance to pythagorean theorem

sbarstys:


cgiger00:
will re-open if the new obstacle interface flops

PR #925, opened 2022-02-16
[nav] Changed driving power

zwgold:
Changes from 2/16 testing, made sure that driving forward is always full power and that the 25% reduction of power is removed.


cgiger00 (approved):
lgtm

PR #935, opened 2022-02-20
Added calibration fields to IMUData struct + IMU now sending magnetometer data

rbridges12:


cgiger00 (approved):
looks good to me, communicate this w cam to make sure it doesn't break any of teleop's code tho

PR #938, opened 2022-02-23
Python Search Point Generation for Spiral-out Search

Dstefanescu2002:
- Removed old method of generating spiral out search points, and replaced it with a new method which reads pre-generated points relative to the rover's starting location from a txt file.
- Added a python file for generating said points, which can be customized to add more search types and can be used to visualize search patterns.
- Changed the rover configuration to hold a file path for the txt file.


jetson/nav/search/spiralOutSearch.cpp
```diff
@@ -3,41 +3,19 @@
...
-            mSearchPointMultiplier.second < 0 ? --mSearchPointMultiplier.second : ++mSearchPointMultiplier.second;
-
-        }
+    ifstream coordinate_file (roverConfig[ "search" ][ "spiralSearchPoints" ].GetString());
+    float rho, phi;
+    while (coordinate_file >> rho >> phi) {
+        Odometry nextSearchPoint = createOdom(rover->roverStatus().path().front().odom, phi, rho, rover);
+        mSearchPoints.push_back( nextSearchPoint );
     }
     insertIntermediatePoints();
```
zwgold, line 20: Make sure to close file stream at the end

zwgold:
Do we want to also make a README/short text document explaining how to use the python script?

jetson/nav/search/spiral_search_points.txt
```diff
@@ -0,0 +1,14 @@
+0.0 90.0
```
cgiger00, line 1: add a comment in the read me for this saying what the numbers in this file mean (like what the colomns are)

jetson/nav/searchPointGeneration.py
```diff
@@ -0,0 +1,78 @@
...
+    coords = generateSquareSpiral(13, 3)
+else:
+    print ("Not a valid type")
+    exit(1)
+show_coords = input("Do you want to display the search coordinates? (y/n)")
+if show_coords == 'y':
+    showCoords(coords)
+
+with open('jetson/nav/search/spiral_search_points.txt', 'w') as f:
+    # print (len(coords), file=f)
```
cgiger00, line 74: is it possible to clear the file before writing to it to avoid a case where a new search has less points than an old one

cgiger00 (changes requested):

jetson/nav/searchPointGeneration.py
Dstefanescu2002: Python will automatically clear the file before writing with the 'w' parameter.

jetson/nav/searchPointGeneration.py
```diff
@@ -0,0 +1,78 @@
...
+    rho = np.sqrt(x**2 + y**2)
+    phi = -1 * (np.arctan2(y, x) * 180 / math.pi - 90)
+    return(rho, phi)
+
+print("\n-----Search types-----\nRadially Equidistant Spiral: 0\nPoint Equidistant Spiral: 1\nSquare Spiral: 2\n")
+search_type = input("Select a search type: ")
+if search_type == '0': # Point Equidistant Spiral
+    # generateSpiralSearchPoints 
+    #           (Radius of spiral, Number of Points, Number of coils)
+    coords = generateSpiralSearchPoints(20, 200, 10)
```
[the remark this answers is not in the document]
Dstefanescu2002: Yes, but it would mean that we would need to standardize the parameters to the different searches. Otherwise, we can also just ask for input depending on the search type that is chosen.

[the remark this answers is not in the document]
Dstefanescu2002: @cgiger00 Since we have multiple search types, there are multiple types of search paths that can be generated. Each of these has different parameters so it is a little hard to take the parameters as inputs to the program. One possible solution would be to prompt the user to specify these parameters depending on the search type that they choose. Otherwise we could specify locations to change these parameters within the file itself. Do you have any way which you would prefer?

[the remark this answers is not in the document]
cgiger00: yes i think prompting the user will be the cleanest way to go about this

jetson/nav/searchPointGeneration.py
```diff
@@ -53,15 +53,17 @@ def cart2pol(x, y):
 
 print("\n-----Search types-----\nRadially Equidistant Spiral: 0\nPoint Equidistant Spiral: 1\nSquare Spiral: 2\n")
 search_type = input("Select a search type: ")
-if search_type == '0': # Point Equidistant Spiral
+if search_type == '0': # POINT EQUIDISTANCE SPIRAL
```
cgiger00, line 56: could this also be just changed to an enum? wary of magic numbers

cgiger00 (commented):
mostly good except for changing the search types/search parameters to be inputs

zwgold (approved):

zwgold (approved):

cgiger00 (approved):

PR #943, opened 2022-03-09
[sim] 2D Simulator Field of View Visualization

danieljiang520:


simulators/nav/src/components/NavSimulator.vue
```diff
@@ -437,29 +443,29 @@ export default class NavSimulator extends Vue {
 
     /* Set up publishing LCMs */
     this.intervalLcmPublish = window.setInterval(() => {
-      this.publish('/auton', { type: 'AutonState', is_auton: this.autonOn });
+      this.publish('/auton', { type: 'AutonState', is_auton: this.autonOn }, false);
```
cgiger00, line 446: why is this set to false?

simulators/nav/src/store/modules/simulatorState.ts
```diff
@@ -115,6 +118,8 @@ const getters = {
 
   roverPath: (simState:SimulatorState):Odom[] => simState.path,
 
+  FOVAreaPath: (simState:SimulatorState):Path2D => simState.FOVAreaPath,
```
cgiger00, line 121: is this something that can be set on the simulator gui?

cgiger00 (commented):
mostly looks good, left a couple comments mostly just questions

simulators/nav/src/components/NavSimulator.vue
danieljiang520: This is just a selector for the Enable LCM function. True if we want to always publish the channel regardless of the state of Enable LCM

simulators/nav/src/store/modules/simulatorState.ts
danieljiang520: This contains the area on the map that the rover's FOV has seen. Do you mean if we can set this area pre-simulation?

cgiger00: yeah that's what i mean, can how wide the FOV and depth of it be changed?

danieljiang520: Yep

https://user-images.githubusercontent.com/71047773/157580287-e600bd19-d701-49ae-aaed-93a1e92724f4.mov

cgiger00 (approved):

zwgold (approved):
lgtm

PR #953, opened 2022-03-14
[nav] Gate Distance Check (3m) 

### LINKED ISSUE #904: [nav] Check post distances at < 3m
Once we are in the gate algorithm part of our code, we need to make sure once we see both posts, if they are indeed within 3 meters of each other to avoid false positives

Delaminer:
resolves #904

This calculates the distance between the two posts. If it detects that they are over 3m apart, output is printed and the robot executes GateSpin.

### DIFF
--- a/jetson/nav/gate_search/gateStateMachine.hpp
+++ b/jetson/nav/gate_search/gateStateMachine.hpp
@@ -57,13 +57,18 @@ private:
 
     NavState executeGateDriveToFarPost();
 
-    NavState executeGateTurnToGateCenter();;
+    NavState executeGateTurnToGateCenter();
+    
+    NavState executeGateShimmy();
 
     NavState executeGateDriveThrough();
 
     void updatePost2Info();
 
     void calcCenterPoint();
+    
+    // Helper function to find the distance between the posts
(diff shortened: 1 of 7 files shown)


rover_msgs/RepeaterDrop.lcm
```diff
@@ -0,0 +1,4 @@
+package rover_msgs;
+
+struct RepeaterDrop {
```
cgiger00, line 3: why is this added back in?

jetson/nav/gate_search/gateStateMachine.hpp
```diff
@@ -57,13 +57,18 @@ class GateStateMachine
 
     NavState executeGateDriveToFarPost();
 
-    NavState executeGateTurnToGateCenter();;
+    NavState executeGateTurnToGateCenter();
+    
+    NavState executeGateShimmy();
```
cgiger00, line 62: wasn't the gateshimmy old logic?

cgiger00 (commented):
is this part of the new gate logic..?

cgiger00:
ok im going to close this PR since we are re-doing the gate logic

PR #956, opened 2022-03-15
[esw] drive code fix for unexpected ACK issue

cgiger00:
Cleaned up threading code so that if it throws an error while a lock is acquired because the odrive is unplugged the lock will still be released


tabiosg (approved):
code is far more readable now, looks good!

PR #963, opened 2022-03-17
Added TargetList to auton data logging

jnnanni:


CameronTressler (approved):

base_station/gui/src/components/AutonTask.vue
```diff
@@ -121,10 +121,8 @@ export default {
         distance: 0
       },
 
-      TargetList: {
-        targetList: [{bearing: 0, distance: 0, id: 0},
-        {bearing: 0, distance: 0, id: 0}]
-      },
+      TargetList: [{bearing: 0, distance: 0, id: 0},
+      {bearing: 0, distance: 0, id: 0}],
```
niwhsa9, line 125: default distance needs to be set to -1

niwhsa9 (changes requested):

base_station/gui/src/components/AutonTask.vue
CameronTressler: I mean we can do that but has having it be 0 in the past been causing issues?

niwhsa9: @zwgold does state machine disregard 0 the same way it disregards -1?

niwhsa9 (approved):

PR #966, opened 2022-03-19
[sim] Merge 2dsim formatting

michmli:


michmli:
Screen Shot 2022-03-19 at 3 56 41 PM

PR #972, opened 2022-03-20
[auton] Navigation New Architecture & Refactoring

qhdwight:
Following: https://github.com/mmartin4972/mrover-workspace/blob/navDesign/NewNavDesign.md
- Use smart pointers instead of raw
- Fix some minor UB with relation to field instantiation
- Separate external rover state into Environment class instead of holding it directly inside the state machine
- ^^ Same as above also with course, in CourseProgress


qhdwight:
Right now it is close to compiling but I would like some feedback on resolving the last couple problems @niwhsa9

jetson/nav/utilities.cpp
```diff
@@ -3,165 +3,135 @@
...
-{
-    return ( PI / 180 ) * ( degree + minute / 60 );
+double degreeToRadian(const double degree, const double minute) {
+    return (PI / 180) * (degree + minute / 60);
 } // degreeToRadian
 
 // Converts the input radians to degrees.
-double radianToDegree( const double radian )
-{
+double radianToDegree(const double radian) {
```
mmartin4972, line 11: You removed the const for this in the header file

jetson/nav/utilities.cpp
```diff
@@ -3,165 +3,135 @@
...
 // // Calculates the modulo of degree with the given modulus.
-double mod( const double degree, const int modulus )
-{
-    double mod = fmod( degree, modulus );
-    if( mod < 0 )
-    {
-        return ( mod + modulus );
-    }
-    return mod;
+double mod(const double degree, const int modulus) {
```
mmartin4972, line 80: you removed the const for this in the header file

jetson/nav/utilities.cpp
```diff
@@ -3,165 +3,135 @@
 #include <cmath>
 
 // Coverts the input degree (and optional minute) to radians.
-double degreeToRadian( const double degree, const double minute )
-{
-    return ( PI / 180 ) * ( degree + minute / 60 );
+double degreeToRadian(const double degree, const double minute) {
```
mmartin4972, line 6: you removed the const for this in the header file

mmartin4972 (commented):
The changes you made look good.

I'm concerned however that you've made all these changes but don't have the code compiling. Because we don't want to mess anything up like Carmen said you need to break these changes apart, apply them individually, then run the unit tests on them. The reason we wanted to make the testing script first was so that we can test all of these changes incrementally. Please break it apart into 5 changes and unit test each individually:
1. moving smart pointers to pointers
2. seems like you ran some auto formatting tool, so have a step for that
3. looked like you refactored a few functions, so make the manual refactoring of these its own step
4. add the environment class modifying only the necessary LCM callbacks
5. add the course class and remove the old LCM callbacks

I'd like to see you make each of these stages of changes then test them. It will be easier for us to review if you break your commits down into this way. It will also make us more confident that no logic was changed since we did everything incrementally.

jetson/nav/utilities.cpp
qhdwight: https://stackoverflow.com/questions/51353648/const-qualification-of-parameters-in-function-declaration

qhdwight:
Will address above for for viewing changes use the whitespace-cognizant mode on GitHub: https://github.com/umrover/mrover-workspace/pull/972/files?diff=unified&w=1

It is much easier to tell what changed semantically

jetson/nav/utilities.cpp
qhdwight: https://stackoverflow.com/questions/51353648/const-qualification-of-parameters-in-function-declaration

niwhsa9: seems like you'd want it in the header since thats the interface specification for other programmers though...

niwhsa9:
I agree with Matt about breaking this up into a couple PRs. I will hold off on my review until then

qhdwight:
Removed logic of maintaining an old and new rover state, removed the actual rover status class and just moved it into rover, decoupled further by extracting rover out of state machine

Still some funky behavior when it reaches it's first waypoint we have to iron out.

qhdwight:
Merged with newest auton-integration

zwgold (approved):
works well on sim and on the rover

jetson/nav/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
+cmake_minimum_required(VERSION 3.8)
```
cgiger00, line 1: can jarvis still build nav?

jetson/nav/layer2_test.cpp
```diff
@@ -1,24 +1,24 @@
...
-    using namespace std::chrono_literals;
-    std::cout << "Testing layer2..." << std::endl;
-   
-    /*Auton::System sys;
-    lcm::LCM lcm_;
-    Layer2 layer2(sys, lcm_);*/
-
-    return 0;   //std::cout << layer2.state << std::endl;
-}
+////layer2 Test
```
cgiger00, line 1: can this be in a .gitignore?

ansible/roles/jetson_dev/tasks/main.yml
```diff
@@ -22,4 +22,5 @@
       - libgirepository1.0-dev
       - gfortran
       - xdotool
+      - libeigen3-dev
```
cgiger00, line 25: why is this in this PR?

jetson/nav/rover.hpp
```diff
@@ -11,15 +11,17 @@
...
 #include "rover_msgs/TargetList.hpp"
 #include "rover_msgs/Waypoint.hpp"
+#include "rover_msgs/AutonDriveControl.hpp"
 #include "rapidjson/document.h"
+#include "courseProgress.hpp"
+#include "environment.hpp"
 #include "pid.hpp"
 
-using namespace rover_msgs;
 using namespace std;
```
cgiger00, line 20: can you get rid of this & just add the specific members of std that you need? like using std::vector? etc

jetson/nav/rover.hpp
```diff
@@ -63,178 +65,118 @@ enum class NavState
...
+    // The rover's current path. The path is initially the same as
+    // the rover's course, however, as waypoints are visited, the
+    // are removed from the path but not the course.
+
+    // The rover's current odometry information.
+    Odometry mOdometry{};
+
+    // The rover's current target information from computer
+    // vision.
+    Target mTargetLeft{-1.0, 0.0, 0};
```
cgiger00, line 167: is -1.0, 0.0, 0 the default? can this just be the default initialization for Target?

jetson/nav/CMakeLists.txt
qhdwight: Yes it builds on both. I can remove cmake if you'd like to keep things clean though

pip_deps/jetson_requirements.txt
```diff
@@ -1,2 +1,2 @@
 numpy==1.16.2
-odrive==0.5.4
+odrive==0.5.1
```
cgiger00, line 2: interesting that this change is in this PR?

jetson/nav/layer2_test.cpp
qhdwight: it should be deleted all together actually

jetson/nav/rover.hpp
qhdwight: sure

jetson/nav/rover.hpp
qhdwight: It's an lcm auto generated, is that possible?

jetson/nav/rover.hpp
qhdwight: done

jetson/nav/CMakeLists.txt
cgiger00: i think it's fine to keep in, can you add a comment in the readme or smth explaining why both are there/when one would the cmake file

jetson/nav/rover.hpp
cgiger00: wym lcm auto generated?

jetson/nav/.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
```
cgiger00, line 1: wait what are the .idea files?

qhdwight: Oops, removed, must've been snuck in by CLion

cgiger00 (commented):
resolsved

jetson/nav/rover.hpp
qhdwight: Target.hpp is auto generated by LCM, so I can't add a ctor I don't think that would set  it to -1.0, 0.0, 0

cgiger00 (approved):
ok lgtm to merge

PR #977, opened 2022-03-24
[teleop] Changed Science GUI from A/B/C to displaying 0/1/2

ErikLytle:
Updated the following files that used A/B/C with underlying 2/1/0 values to 0/1/2:
- StripTest.vue (Strip Test Controls)
- Amino.vue (Heater controls)
- Carousel.vue
- Chlorophyll.vue (Chlorophyll Spectral data table)

### DIFF
--- a/base_station/gui/src/components/Amino.vue
+++ b/base_station/gui/src/components/Amino.vue
@@ -6,37 +6,37 @@
   <div class="box1">
     <label for="site">Choose a site:</label>
     <select v-model = "site" name="site" id="site">
-      <option value="2">Site A</option>
-      <option value="1">Site B</option>
-      <option value="0">Site C</option>
+      <option value="0">Site 0</option>
+      <option value="1">Site 1</option>
+      <option value="2">Site 2</option>
     </select>
   </div>
   <div class="box1" v-if="site == 2">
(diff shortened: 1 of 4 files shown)


base_station/gui/src/components/Amino.vue
```diff
@@ -6,37 +6,37 @@
...
     <select v-model = "site" name="site" id="site">
-      <option value="2">Site A</option>
-      <option value="1">Site B</option>
-      <option value="0">Site C</option>
+      <option value="0">Site 0</option>
+      <option value="1">Site 1</option>
+      <option value="2">Site 2</option>
     </select>
   </div>
   <div class="box1" v-if="site == 2">
```
tabiosg, line 14: could probably rearrange this so that site 0 comes in first, doesn't really matter though

tabiosg (approved):
looks good, merge it in whenever u want to. can rearrange code so it's in numeric order if you want before then though

CameronTressler:
@ErikLytle  I guess one thing I failed to mention earlier: ints are better than strings because we can easily use them to index into arrays. Some code duplication could be removed now if we used arrays more but I don't think it's currently worth the time.

PR #979, opened 2022-03-24
Spiral out search pattern upgrades

Dstefanescu2002:
- Decreased number of search points for spiral out search by increasing distance between spiral lines. This was tested with the new simulator feature that shows search area to ensure that we are still searching the full region.
- Added spiral-in search pattern capabilities to the python search point generation code. This is not include in the current search pattern.


cgiger00:
can you post an image of the coverage this new spiral gives from the simulator viewer?

Dstefanescu2002:
> can you post an image of the coverage this new spiral gives from the simulator viewer?
image
The area that is not search in the middle is within a five meter radius of the point, so the AR tag will not be there.

zwgold:
The pattern looks solid, plus at any rate, driving to that point should cover most of that area.

zwgold (approved):
looks p solid, coverage good

PR #982, opened 2022-03-26
[nav] Remove intermediate, include change for checking if immediately at point

zwgold:
Removes the intermediate points from the search, function is still in there just in case. Also, checks if we are immediately at the first point


niwhsa9 (approved):
lgtm

PR #986, opened 2022-03-27
[teleop] Add tangent arrow

CameronTressler:


CameronTressler:
Sorry I forgot to PR this Friday

rbridges12 (approved):
looks good to me

base_station/gui/src/components/PlaybackAutonData.vue
```diff
@@ -78,6 +80,9 @@ export default {
               case 'Odom bearing':
```
cgiger00, line 80: nit pick, could these be enums instead of strings?

base_station/gui/src/store/modules/autonomy.js
```diff
@@ -9,8 +9,7 @@ const state = {
   playbackLat: [],
   playbackLon: [],
   playbackBearing: [],
```
cgiger00, line 11: can you add a distinction of which type of bearing this is? like is this the rover's bearing (from the imu?) so it's clear how its different from gpsBearing

cgiger00 (changes requested):

base_station/gui/src/components/PlaybackAutonData.vue
CameronTressler: No, they're strings from the csv file

base_station/gui/src/store/modules/autonomy.js
CameronTressler: Anything that isn't labeled refers to the rover odometry, but I'll add that

CameronTressler:
image

cgiger00 (approved):

PR #988, opened 2022-03-27
GPS encoder stop, filter crash fix

rbridges12:
Added fix to the filter to stop updating state when encoders read zero velocity, changed filter to send no odometry when GPS is unplugged instead of crashing.


RyanBoring (approved):

cgiger00 (commented):
did you see how this behaves if you're stopped then start moving again?

rbridges12:
Yeah, I tested a bunch of combinations of stopping, starting, and unplugging the GPS/IMU

cgiger00:
ok lgtm then

PR #989, opened 2022-03-27
Additional checks regarding control state in IK

bli76:
- Automatically set state to closed-loop when we tell the arm to move
- Safety check if closed-loop was aborted or other unexpected state is encountered before updating control state

### DIFF
--- a/jetson/ra_kinematics/mrover_arm.cpp
+++ b/jetson/ra_kinematics/mrover_arm.cpp
@@ -179,7 +179,6 @@ void MRoverArm::target_orientation_callback(std::string channel, TargetOrientati
         std::cout << "Received target but not currently waiting for target.\n";
         return;
     }
-    control_state = ControlState::CALCULATING;
 
     std::cout << "Received target!\n";
     std::cout << "Target position: " << msg.x << "\t" << msg.y << "\t" << msg.z << "\n";
@@ -203,9 +202,21 @@ void MRoverArm::target_orientation_callback(std::string channel, TargetOrientati
         // send popup message to GUI
         lcm_.publish("/debug_message", &msg);
 
-        control_state = ControlState::WAITING_FOR_TARGET;
         return;
     }
+
+    if (control_state != ControlState::WAITING_FOR_TARGET) {
+        std::cout << "IK calculations cancelled\n";
+
+        if (control_state != ControlState::OFF) {
+            std::cout << "UNEXPECTED BEHAVIOR: cancelled IK but did not leave closed-loop! Setting state to off.\n";
+            control_state = ControlState::OFF;
+        }
+
+        return;
+    }
+        
+    control_state = ControlState::CALCULATING;
 
     Vector6d point;
     point(0) = (double) msg.x;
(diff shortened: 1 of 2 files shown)


cgiger00 (approved):

cgiger00:
lgtm, cam probably has more input tho

jetson/ra_kinematics/mrover_arm.cpp
```diff
@@ -209,10 +202,22 @@ void MRoverArm::target_orientation_callback(std::string channel, TargetOrientati
...
+    if (control_state != ControlState::WAITING_FOR_TARGET) {
+        std::cout << "IK calculations cancelled\n";
+
+        if (control_state != ControlState::OFF) {
+            std::cout << "UNEXPECTED BEHAVIOR: cancelled IK but did not leave closed-loop! Setting state to off.\n";
+            control_state = ControlState::OFF;
+        }
+
+        return;
+    }
```
CameronTressler, line 217: Any way this could become a function as well?

CameronTressler:
Didn't think of this till now but does it make sense to have OFF and WAITING_FOR_TARGET be different states?

jetson/ra_kinematics/mrover_arm.cpp
```diff
@@ -502,6 +528,10 @@ void MRoverArm::zero_position_callback(std::string channel, Signal msg) {
 }
 
 void MRoverArm::arm_adjust_callback(std::string channel, ArmAdjustments msg) {
+    if (control_state == ControlState::OFF) {
+        set_to_closed_loop();
+    }
+
```
CameronTressler, line 534: Honestly I don't think this is needed, and we can also remove the 4 lines right below this. We're about to call target_orientation, which should do both of these checks for us. If you disagree lmk

bli76:
> Didn't think of this till now but does it make sense to have OFF and WAITING_FOR_TARGET be different states?

Guess not, especially now that we automatically switch to closed-loop if it was off.

CameronTressler:
> > Didn't think of this till now but does it make sense to have OFF and WAITING_FOR_TARGET be different states?
> 
> Guess not, especially now that we automatically switch to closed-loop if it was off.

Okay I guess let’s leave as is for now, I don’t think it’s worth the potential to break something that isn’t broken

CameronTressler:
Did you run 6 joint arm in the sim to test behavior of this since your last changes?

bli76:
> Did you run 6 joint arm in the sim to test behavior of this since your last changes?

Yep, everything seems to work fine including shutting off closed-loop midway.

bli76:
> > > Didn't think of this till now but does it make sense to have OFF and WAITING_FOR_TARGET be different states?
> > 
> > 
> > Guess not, especially now that we automatically switch to closed-loop if it was off.
> 
> Okay I guess let’s leave as is for now, I don’t think it’s worth the potential to break something that isn’t broken

Actually I took another look at the code and I think we need the separate states to know when to update the Closed Loop button on the GUI

CameronTressler:
> > Did you run 6 joint arm in the sim to test behavior of this since your last changes?
> 
> Yep, everything seems to work fine including shutting off closed-loop midway.

Lit

CameronTressler:
> > > Didn't think of this till now but does it make sense to have OFF and WAITING_FOR_TARGET be different states?
> > 
> > 
> > Guess not, especially now that we automatically switch to closed-loop if it was off.
> 
> Actually I took another look at the code and I think we need the separate states to know when to update the Closed Loop button on the GUI

Ah yeah you're right.

PR #997, opened 2022-03-31
[teleop] Added an auto shutdown timer for UV bulb

catkims:
image


base_station/gui/src/components/Chlorophyll.vue
```diff
@@ -19,8 +19,18 @@
       <span v-if="UVLED == 0" class="toggle__label" >UV LEDs Off</span>
 
       <input type="checkbox" id="toggle_button">
-      <span class="toggle__switch" v-if="UVLED == 0" v-on:click="UVLED=1, setPart(mosfetIDs.UVLED, true)"></span>
-      <span class="toggle__switch" v-if="UVLED == 1" v-on:click="UVLED=0, setPart(mosfetIDs.UVLED, false)"></span>
+      <span class="toggle__switch" v-if="UVLED == 0" v-on:click="UVLED=1, UVshutdown(true)"></span>
+      <span class="toggle__switch" v-if="UVLED == 1" v-on:click="UVLED=0, UVshutdown(false)"></span>
```
CameronTressler, line 23: The above two lines could be condensed to one line if we use the value of UVLED in the on:click= part

base_station/gui/src/components/Chlorophyll.vue
```diff
@@ -19,8 +19,18 @@
...
+    </label>
+  </div>
+  <div class="shutdown">
+    <label for="toggle_button" :class="{'active': shutdown == 1}" class="toggle__button">
+      <span v-if="shutdown == 1" class="toggle__label" >UV Auto shutoff On</span>
+      <span v-if="shutdown == 0" class="toggle__label" >UV Auto shutoff Off</span>
+
+      <input type="checkbox" id="toggle_button">
+      <span class="toggle__switch" v-if="shutdown == 0" v-on:click="shutdown=1"></span>
+      <span class="toggle__switch" v-if="shutdown == 1" v-on:click="shutdown=0"></span>
```
CameronTressler, line 33: Similarly this code could be condensed

CameronTressler (changes requested):

CameronTressler:
My understanding was that the Scoop UV bulb should be set on a timer. Not the Chlorophyll UV. You should check with Gabe which he wants.

Sorry I didn't realize this confusion earlier when we were looking at this together

cgiger00:
im under that impression too --it's the scoop UV, can this PR be turned into a draft in the meantime while that's being looked into?

jnnanni (approved):
Lg2m

base_station/gui/src/components/ScoopUV.vue
```diff
@@ -102,6 +117,15 @@ export default {
         'device': id,
         'enable': enabled
       })
+    },
+    UVshutdown: function(status) {
+      this.setPart(this.mosfetIDs.uvBulb, status);
+      if (this.shutdown === 1) {
```
CameronTressler, line 123: So looking at this more, my concern is we could turn it on, turn it off, turn it on, and then the first timeout would turn it off in shorter than 2 minutes. Can you use the cleartimeout function to mitigate this?
https://developer.mozilla.org/en-US/docs/Web/API/clearTimeout

CameronTressler (changes requested):

CameronTressler (approved):
lgtm

cgiger00 (approved):

PR #998, opened 2022-03-31
[Auton-GUI] Added intermediate step between toggling auton mode

jnnanni:
image
Button will now turn yellow when toggling auton mode on or off and indicated which it is trying to be set to. After receiving a message with the proper nav status the button will turn red/green. Descriptions of behavior below.
Off -> On: the button will turn yellow and read "Setting to true", then turn green once a nav status other than "Off" is received.
On -> Off: the button will turn yellow and read "Setting to false", then turn red once an "Off" nav status is received.
Note that the button can still be clicked again while in a yellow state and it will switch to the other transition state.

### DIFF
--- a/base_station/gui/src/components/WaypointEditor.vue
+++ b/base_station/gui/src/components/WaypointEditor.vue
@@ -42,7 +42,7 @@
     <div class="col-wrap" style="left: 50%">
       <div class="box datagrid">
         <div class="autonmode">
-          <Checkbox ref="checkbox" v-bind:name="'Autonomy Mode'" v-on:toggle="toggleAutonMode($event)"/>
+          <Checkbox ref="checkbox" v-bind:name="autonButtonText" v-bind:color="autonButtonColor"  v-on:toggle="toggleAutonMode($event)"/>
         </div>
         <div class="stats">
           <p>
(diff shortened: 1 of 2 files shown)


base_station/gui/src/components/WaypointEditor.vue
```diff
@@ -128,6 +132,16 @@ export default {
 
     this.$parent.subscribe('/nav_status', (msg) => {
       this.nav_status = msg
+      if(this.waitingForNav){
+        if(this.nav_status.nav_state_name!="Off" && this.autonEnabled){
+          this.waitingForNav = false
+          this.autonButtonColor = "green"
+        }
+        else if(this.nav_status.nav_state_name=="Off" && !this.autonEnabled){
```
cgiger00, line 140: @zwgold have you fixed the bug where nav doesn't go to off when auton is disabled?

zwgold: @cgiger00 Refer to #1004 for the fix to make sure we go to off if auton is disabled

cgiger00:
lgtm

cgiger00 (approved):

PR #999, opened 2022-04-02
Encoder stop fix

rbridges12:
Changed encoder stopping to detect any non zero encoder velocity, not just positive velocities.


cgiger00:
lgtm

cgiger00 (approved):

PR #1000, opened 2022-04-02
[esw] Moved polarity flip to esw as a hardware change since we will never be driving backwards

Polishdudealan:


jetson/odrive_bridge/src/__main__.py
```diff
@@ -329,7 +329,7 @@ def drive_vel_cmd_callback(channel, msg):
             global left_speed, right_speed
 
             speed_lock.acquire()
-            left_speed, right_speed = cmd.left, cmd.right
+            left_speed, right_speed = -cmd.left, -cmd.right,
```
cgiger00, line 332: can you leave this as cmd.left and cmd.right and just flip the negative sign in modrive.set_vel?

cgiger00 (changes requested):

CameronTressler:
I think I'm missing something, why are we PR-ing this into auton-integration and not auton-drive-change?

cgiger00:
they're both being PRed into auton-integration, doesn't rly matter either way imo

tabiosg (approved):

ankithu (approved):

CameronTressler (approved):

cgiger00 (approved):

PR #1002, opened 2022-04-02
[Auton, Teleop] Auton drive change

CameronTressler:


jetson/teleop/src/__main__.py
```diff
@@ -86,66 +87,68 @@ def deadzone(magnitude, threshold):
...
+
+        # Convert arcade drive to tank drive
+        # TODO: possibly flip signs of rotation if necessary
+        angular_op = (angular / 2) / (abs(linear) + 0.5)
+        vel_left = linear + angular_op
+        vel_right = linear - angular_op
+
+        # Account for reverse
+        if self.reverse:
+            tmp = vel_left
```
cgiger00, line 122: (for in the future) i dont think this produces a true polarity flip, like it flips turn + driving at the same time bc they're both positive i think, but for turning in place like where left turn is like L=-1, R=1, like this 'flips' it to L=-1 and R=1 which when the polarity is flipped is a right turn?

nav_output.txt
```diff
@@ -0,0 +1,1006 @@
+stopping
```
cgiger00, line 1: can this be added to a .gitignore?

cgiger00 (changes requested):

jetson/teleop/src/__main__.py
CameronTressler: I'm not following. Turning left in place is the same for both directions -- it's turning counterclockwise.

nav_output.txt
ankithu: I'll just delete these files, they're not important

jetson/nav/rover.cpp
```diff
@@ -167,6 +167,9 @@ DriveStatus Rover::drive( const Odometry& destination )
 // on-course or off-course.
 DriveStatus Rover::drive( const double distance, const double bearing, const bool target )
 {
+    //if (target){
```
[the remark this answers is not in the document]
ankithu: yea, I'll put in another PR later that does the debug messages stuff if that's alright.

jetson/nav/rover.cpp
```diff
@@ -229,19 +239,24 @@ bool Rover::turn( double bearing )
...
     }
     double turningEffort = mBearingPid.update( mRoverStatus.odometry().bearing_deg, bearing );
+    std::cout << "cur bearing: " << mRoverStatus.odometry().bearing_deg << " target bearing: " << bearing << " effort: " << turningEffort << std::endl;
     double minTurningEffort = mRoverConfig[ "navThresholds" ][ "minTurningEffort" ].GetDouble() * ( turningEffort < 0 ? -1 : 1 );
     if( isTurningAroundObstacle( mRoverStatus.currentState() ) && fabs( turningEffort ) < minTurningEffort )
     {
         turningEffort = minTurningEffort;
     }
-    publishJoystick( 0, turningEffort, false );
+    double left_vel = max(min(1.0, turningEffort), -1.0);
```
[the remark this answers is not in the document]
ankithu: resolved in recent commit

config/nav/config.json
```diff
@@ -56,6 +56,7 @@
 		"repeaterDropInitChannel": "/rr_drop_init",
 		"repeaterDropCompleteChannel": "/rr_drop_complete",
 		"joystickChannel": "/autonomous",
```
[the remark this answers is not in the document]
ankithu: resolved in commit

jetson/nav/rover.cpp
```diff
@@ -197,14 +199,15 @@ DriveStatus Rover::drive( const double distance, const double bearing, const boo
 // does not calculate if you have arrive at a specific location and
 // this must be handled outside of this function.
 // The input bearing is an absolute bearing.
+//TODO: I'm 90% sure this function is redundant, we should just remove it
```
cgiger00, line 202: lol

jetson/teleop/src/__main__.py
cgiger00: but if you're flipped that's a right turn

cgiger00: we can talk about it offline since it's not critical to this PR, but i think this explains why in testing we noticed even from the joystick curve turning to the left one way would make the rover go left but a left turn in place would make it go right

jetson/nav/rover.cpp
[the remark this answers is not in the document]
cgiger00: sg

config/nav/config.json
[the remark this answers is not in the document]
cgiger00: oh, i meant deleting publishing to /autonomous

cgiger00 (approved):
good enough for testing, i would rather you delete publishing to /autonomous before merging but im just going to approve it

PR #1004, opened 2022-04-03
[nav] Target in Drive + Other Fixes

zwgold:
This PR addresses the following:

- Changes the distance threshold for reaching a target to 1.5 meters per our talk on trying to place the ZED for obstacles and how that would impact Targets
- Adds three extra spiral points (17 total) to ensure we get 20 meters on all four corners
- Makes sure when nav is turned off in the `base_station/gui` or `simulators/nav` that we go to "Off" instead of staying in "Done"
- Checks if Target is the correct distance away from a given waypoint (at most 20 meters) and it's ID (to make sure we don't try and reach the wrong one)


jetson/nav/stateMachine.cpp
```diff
@@ -270,10 +270,6 @@ NavState StateMachine::executeOff()
         mCompletedWaypoints = 0;
         mTotalWaypoints = mRover->roverStatus().course().num_waypoints;
 
-        if( !mTotalWaypoints )
-        {
-            return NavState::Done;
-        }
         return NavState::Turn;
```
cgiger00, line 273: why does execute off return turn?

cgiger00 (approved):

jetson/nav/stateMachine.cpp
zwgold: This part of `executeOff` is checking if the base station sends us the `is_auton` boolean, so within that branch it'll check if it has to move or not. The comment above should explain it, but basically it should have a transition from off to the first move the rover does (which is turn to face a waypoint)

cgiger00: ok...some of these variables/functions definitely could be renamed

PR #1009, opened 2022-04-03
Create science-bridge and cameras ansible scripts

jcp419:
wowee

### DIFF
--- a/ansible/roles/jetson_services/files/rover-jetson-cameras.service
+++ b/ansible/roles/jetson_services/files/rover-jetson-cameras.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=MRover Cameras
+
+[Service]
+EnvironmentFile=/etc/default/rover.env
+User={{ running_user }}
+ExecStart={{ workspace_dir }}/jarvis exec jetson_cameras_
+Restart=always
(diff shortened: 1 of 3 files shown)


ansible/roles/jetson_services/files/rover-jetson-cameras.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=MRover Cameras
+
+[Service]
+EnvironmentFile=/etc/default/rover.env
+User={{ running_user }}
+ExecStart={{ workspace_dir }}/jarvis exec jetson_cameras_
```
tabiosg, line 7: jetson_cameras, why the extra underscore :^)

tabiosg (changes requested):
You will also need to change the ansible/roles/jetson_services/tasks/main.yml file

tabiosg:
Never mind, main.yml is already updated (I didn't see newest commit). Anyway, fix the extra underscore

ansible/roles/jetson_services/files/rover-jetson-cameras.service
jcp419: why must you tear me down

PR #1010, opened 2022-04-03
Jarvis Launch Changes

mmartin4972:
* Specify the -r flag to run only and not build
* Added the export LCM command, which will be executed anytime launch is called
* Added commands to localization launch

### DIFF
--- a/jarvis_files/jarvis_cmd/src/jarvis/launch.py
+++ b/jarvis_files/jarvis_cmd/src/jarvis/launch.py
@@ -9,10 +9,9 @@ import os
 #
 ########################
 def launch_dir(ctx, package, ssh, only_run, opts):
-    if ssh :
-        wid = launch_terminal()
-        exec_cmd(wid, "export LCM_DEFAULT_URL='udpm://239.255.76.67:7667?ttl=255'", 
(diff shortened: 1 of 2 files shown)


jarvis_files/jarvis_cmd/src/jarvis/launch.py
```diff
@@ -8,70 +8,78 @@
 # Function that parses launch command
 #
 ########################
-def launch_dir(ctx, package, ssh, opts):
+def launch_dir(ctx, package, ssh, only_run, opts):
+    wid = launch_terminal()
+    exec_cmd(wid, "export LCM_DEFAULT_URL='udpm://239.255.76.67:7667?ttl=255'", 
+            ssh=ssh, ip=AUTON_IP, close_terminal = True)
```
cgiger00, line 14: can you add this for drive too?

cgiger00 (changes requested):

jarvis_files/jarvis_cmd/src/jarvis/launch.py
mmartin4972: yes

mmartin4972:
Tested on rover

cgiger00 (approved):
lgtm

PR #1012, opened 2022-04-05
[Teleop] Drive polarity flips, improved controls

CameronTressler:


tabiosg (approved):

jetson/teleop/src/__main__.py
```diff
@@ -114,8 +114,8 @@ def teleop_drive_callback(self, channel, msg):
         # Convert arcade drive to tank drive
         # TODO: possibly flip signs of rotation if necessary
         angular_op = (angular / 2) / (abs(linear) + 0.5)
-        vel_left = linear + angular_op
-        vel_right = linear - angular_op
+        vel_left = linear - angular_op
```
cgiger00, line 117: wait why did you flip this?

base_station/gui/src/components/DriveControls.vue
```diff
@@ -68,12 +69,24 @@ export default {
...
+              else if (rotation < -1) {
+                rotation = -1
+              }
 
               const joystickData = {
                 'type': 'Joystick',
-                'forward_back': gamepad.axes[JOYSTICK_CONFIG['forward_back']],
-                'left_right': gamepad.axes[JOYSTICK_CONFIG['left_right']],
+                // forward on joystick is -1, so invert
+                'forward_back': -1 * gamepad.axes[JOYSTICK_CONFIG['forward_back']],
```
cgiger00, line 88: i think this double flips now bc of this PR https://github.com/umrover/mrover-workspace/pull/1013, i can revert it back there?

jetson/teleop/src/__main__.py
CameronTressler: Angular should be positive when rotating left. We're going to stick with that for consistency now.

base_station/gui/src/components/DriveControls.vue
CameronTressler: Yeah you can, or I can push a commit changing that later today

CameronTressler: I think the GUI should be in charge of flipping the joystick value so that lcms sent over /drive_control are not reversed

CameronTressler:
With these changes it should send esw positive values for going forward and negative values for going backwards. As long as the other PR (#1013) lines up with that we should be good (we should revert the teleop changes in that PR

base_station/gui/src/components/DriveControls.vue
cgiger00: sg

jetson/teleop/src/__main__.py
cgiger00: does this flip the turn direction though?

CameronTressler: yes

cgiger00: why do we want the turn direction flipped? we got it working w how the turn direction perviously was, if we flip turn here the left and right sides need to change in the esw code too in this PR

CameronTressler: I believe this flip is accounted for in the change in the GUI. This will send 1 for right and -1 for left when we want to turn counterclockwise. It will send both as 1 when we want to drive forward. My understanding was the recent polarity flip in esw would account for this. If not, I think we should change the esw code to match this

cgiger00: the esw polarity flip accounted for the change in the gui already, for testing we just changed the teleop end so that it was no longer sending -1 and -1 for right and left when you push the joystick forward --the turning direction was left unchanged and correct. the polarity flip did not account for the turning also being flipped since that change was not in the teleop code and from my understanding flipping the turning then makes turning backwards. interestingly enough if the esw code is also changed to account for this flip it goes back to what is was originally prior to any polarity flips, so i think teleop has been backwards this entire time then lol

CameronTressler: I think the bottom line is that if ESW moves a side of the drive train "forward" (in the direction we consider to be the front of the rover) when we send a value of positive 1, then these changes should line up and things will work.

I feel like separate packages should be tested and QA'd individually rather than together

CameronTressler: We have the time now that we're not in the middle of testing to make sure that both teleop's and esw's programs are behaving correctly, so I think we may as well make sure both make sense in terms of polarity

cgiger00: we did verify in testing that this worked, so i'm fairly sure if teleop makes this change esw can just revert all the polarity changes they ddi

cgiger00: i think in general when making these sorts of changes you have to consider the other packages it will affect

CameronTressler: Yeah I agree that makes sense, sorry if I wasn't being clear on how these changes would affect things. I added a commit to revert the esw polarity flip from the other day so this should be all good?

PR #1020, opened 2022-04-07
[Teleop] Control improvements

CameronTressler:
Add useful IK buttons to science GUI
Update science presets
Add wrist turn count functionality
Attempt to lock joints that don't move in motion planning


base_station/gui/src/components/ArmControls.vue
```diff
@@ -6,6 +6,9 @@
     <div class="keyboard">
       <GimbalControls/>
     </div>
+    <div style="font-color: abs(wristTurnCount) >= 2 ? red : black;">
```
CameronTressler, line 9: @catkims this doesn't seem to work properly, not sure the syntax to fix

CameronTressler: This shouldn't block a merge though

tabiosg (approved):
Just to make it clear, wrist turn count hasn't been implemented yet on esw's end, hopefully that doesn't break anything

CameronTressler:
> Just to make it clear, wrist turn count hasn't been implemented yet on esw's end, hopefully that doesn't break anything

It shouldn't. Especially not for science arm tomorrow

base_station/gui/src/components/ArmControls.vue
catkims: I'm looking into it now

PR #1024, opened 2022-04-09
[teleop] Made toggle button component, created heater status indicator

ErikLytle:
- Made toggle component, updated all toggles to use the component
- Changed Amino Heater & Auto Shutdown to have a toggle for the intended value, with a status light for actual heater status (enabled)

Screen Shot 2022-04-09 at 4 55 47 PM


base_station/gui/src/components/ScoopUV.vue
```diff
@@ -3,109 +3,52 @@
...
   props: {
     mosfetIDs: {
       type: Object,
       required: true
     }
   },  
   methods: {
+    toggleScoop: function () {
+      this.scoopActive = !this.scoopActive
+      this.setPart(this.mosfetIDs.scoopUV, this.scoopActive)
```
catkims, line 44: uvBulb

CameronTressler (approved):

tabiosg (approved):

tabiosg:
looks good!

tabiosg (approved):

CameronTressler:
@ErikLytle @catkims can one of you resolve the conflict by adding the scoop UV auto shut-off toggle as a component by tomorrow evening?

jnnanni (approved):
lg2m

CameronTressler (changes requested):
I think the toggle component is overly complex. There should ideally just be one variable that we can change that relates to the state, instead of isActive and disabled and defaultState. This would hopefully make it easier to get the toggle to reflect the change in auto shutdown

base_station/gui/src/components/ToggleButton.vue
```diff
@@ -0,0 +1,135 @@
...
+
+        <input type="checkbox" :disabled="disabled" :id="id + '_button'" v-model="checkedValue">
+        <span class="toggle__switch"></span>
+    </label>
+</template>
+
+<script>
+    export default {
+    props: {
+        disabled: {
```
CameronTressler, line 14: I think you can remove disabled. It's redundant with current_state

base_station/gui/src/components/ToggleButton.vue
```diff
@@ -0,0 +1,135 @@
...
+            type: Boolean,
+            default: false
+        },
+        labelEnableText: {
```
CameronTressler, line 18: Can you make the endabled and disabled text props required instead of having defaults?

base_station/gui/src/components/ToggleButton.vue
```diff
@@ -0,0 +1,135 @@
...
+            default: false
+        }
+    },
+    data() {
+        return {
+            currentState: this.defaultState
+        }
+    },
+    watch: {
+        defaultState: function defaultState() {
```
CameronTressler, line 42: I don't think it makes sense to have defaultState be watched. Theoretically this value should never change. However, I think you could currentState a prop passed in from the parent component instead of a local data member. In that case, you could just get rid of defaultState all together. The parent component would set currentState to a variable (like scoopUVActive) which has its own default value and can be changed by the parent

base_station/gui/src/components/ToggleButton.vue
```diff
@@ -0,0 +1,135 @@
...
+            this.currentState = Boolean(this.defaultState)
+        }
+    },
+    computed: {
```
CameronTressler, line 46: checkedValue looks very useful, but I'm not sure the first three values need to be computed rather than just grabbing the value that each of them return directly

PR #1025, opened 2022-04-09
Fix motion planner to keep joints stationary if starting and target angles are the same

bli76:
- Set target angle exactly equal to starting angle if they're close enough
- Change motion planner threshold to match threshold used in above
- Break out of motion planner if it takes too long
- A few missing `std::` scope specifiers

### DIFF
--- a/jetson/ra_kinematics/mrover_arm.cpp
+++ b/jetson/ra_kinematics/mrover_arm.cpp
@@ -289,11 +289,11 @@ void StandardArm::target_orientation_callback(std::string channel, TargetOrienta
     plan_path(hypo_state, goal);
 }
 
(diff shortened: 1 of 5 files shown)


jetson/ra_kinematics/mrover_arm.cpp
```diff
@@ -289,10 +289,11 @@ void StandardArm::target_orientation_callback(std::string channel, TargetOrienta
     plan_path(hypo_state, goal);
 }
 
-void MRoverArm::plan_path(ArmState& hypo_state, const std::vector<double> &goal) {
```
CameronTressler, line 292: You can keep this as const if you just pass original goal into the second call to rrt_connect

CameronTressler (approved):

PR #1027, opened 2022-04-09
[nav] SearchPoints LCM

zwgold:
Nav will send the search points when we are in the search turn and drive modes so that TeleOp can project them on the map when we perform our search spiral.


zwgold:
Right now, the only states that send the paths are searchTurn and searchDrive. Should we send them when we go into obstacle mode or when we locate a target? Or even send them constantly and have TeleOp handle logic when it's empty or in certain states?

zwgold:
Updated to always publish in searchStateMachine when those states within it are running

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -150,6 +151,20 @@ NavState SearchStateMachine::executeDriveToTarget() {
     return NavState::DriveToTarget;
 } // executeDriveToTarget()
 
+// Sends search path rover takes when trying to find posts
+void SearchStateMachine::publishSearchPoints() {
+    // Construct vector from deque
+    std::vector<Odometry> arr(mSearchPoints.begin(), mSearchPoints.end());
+    SearchPoints searchPoints{
+        .search_pattern_size  = (int32_t)mSearchPoints.size(),
```
cgiger00, line 159: in general kind of a nitpick can you use ```std::static_cast<>```

zwgold: yeah lemme fix that, good call

cgiger00 (approved):
lgtm

PR #1028, opened 2022-04-10
[sim] Changed 2dsim Obstacle interface

danieljiang520:


simulators/nav/src/utils/types.ts
```diff
@@ -128,12 +128,20 @@ export interface Obstacle {
...
+  Obstacle LCM. +x is to the right of camera, +y is down from camera,
+  +z is forward (out) from the camera */
+export interface ObstacleNewInterfaceMessage {
+  bottom_left_coordinate_meters:number[];
+  top_right_corrdinate_meters:number[];
+}
 
 /* Interface representing the Obstacle LCM. This must be the same as the
    Obstacle LCM. */
 export interface ObstacleMessage {
```
zwgold, line 141: Do we need this anymore? Or are you keeping it in for reference? (might want to but a deprecated note if you are)

rover_msgs/Obstacle.lcm
```diff
@@ -1,7 +1,6 @@
 package rover_msgs;
 
 struct Obstacle {
-	double bearing; // from straight ahead
-	double rightBearing;
-	double distance; // from straight ahead
-}
+	double bottom_left_coordinate_meters[3];
```
cgiger00, line 4: this will probably break perceptions obstacle code before the new one gets PRed in, @niwhsa9 if obstacle id disabled does the perception code not try to do anything at all (like even subscribe/publish to the obstacle channel, initialize an obstacle object, etc) with this LCM?

simulators/nav/src/.eslintrc.js
```diff
@@ -53,6 +57,7 @@ module.exports = {
         "nav_state_name",
         "num_waypoints",
         "signal_strength",
+        "top_right_corrdinate_meters",
```
cgiger00, line 60: coordinate

simulators/nav/src/components/lcm_center/ObstacleLCM.vue
```diff
@@ -6,8 +6,8 @@
   <div class="box">
     <fieldset class="obstacle">
       <legend>Obstacle</legend>
-      <p>Distance: {{ dist }} m</p>
-      <p>Bearing: {{ bear }}º</p>
+      <p>bottom_left_coordinate_meters: {{ dist }} m</p>
+      <p>top_right_corrdinate_meters: {{ bear }}º</p>
```
cgiger00, line 10: coordinate

rover_msgs/Obstacle.lcm
```diff
@@ -1,7 +1,6 @@
...
+	double top_right_corrdinate_meters[3];
+}
```
cgiger00, line 6: wait doesn't this also need the ObstacleList.lcm from this [ICD](https://docs.google.com/document/d/14QWH8xS4Eq984rMYm_qVhqzoIqVrVtYsluBA5v8_nOQ/edit)?

simulators/nav/src/store/modules/roverState.ts
```diff
@@ -50,7 +50,8 @@ const state:RoverState = {
 
   obstacleMessage: {
     distance: -1,
-    bearing: 0
+    bearing: 0,
+    size: 0.5
```
cgiger00, line 54: what is this the size of?

simulators/nav/src/utils/types.ts
```diff
@@ -128,12 +128,20 @@ export interface Obstacle {
   size:number; /* meters */
 }
 
+/* Type representing the Obstacle LCM. This must be the same as the
+  Obstacle LCM. +x is to the right of camera, +y is down from camera,
+  +z is forward (out) from the camera */
+export interface ObstacleNewInterfaceMessage {
```
cgiger00, line 134: im a bit confused about where the ObstacleList is

simulators/nav/src/utils/utils.ts
```diff
@@ -8,7 +10,9 @@ import {
   OdomFormat,
   Point2D,
   Speeds,
-  ZedGimbalPosition
+  ZedGimbalPosition,
+  ObstacleNewInterfaceMessage,
+  ObstacleMessage
```
cgiger00, line 15: why are there two of these? is one of them the old one? and if so, why are we keeping it?

simulators/nav/src/components/lcm_center/ObstacleLCM.vue
danieljiang520: resolved

simulators/nav/src/utils/utils.ts
danieljiang520: Because I did not want to mess up with how the current 2d sim detects obstacles within the FOV, I basically did a conversion of the old obstacle message (distance, bearing) to its bounding box using the size of the obstacle.

danieljiang520: It would be better to rename the old obstacle message. Will do that later

simulators/nav/src/store/modules/roverState.ts
danieljiang520: The size of an obstacle. It is for calculating its bounding box

simulators/nav/src/utils/utils.ts
danieljiang520: Never mind, I can get rid of the old one entirely after making some changes

simulators/nav/src/utils/types.ts
danieljiang520: Because I did not want to mess up with how the current 2d sim detects obstacles within the FOV, I basically did a conversion of the old obstacle message to the new one and calculate the bounding box using the size of the obstacles

PR #1031, opened 2022-04-10
[teleop] Receive and show search points in GUI

CameronTressler:


CameronTressler:
image

CameronTressler:
Will automatically update path when auton sends something along /search_points

cgiger00 (approved):

CameronTressler:
> What was the purpose of the changes to WaypointEditor?

The select function wasn’t doing really anything except logging errors on the console. And we should always be keeping track of longitude in terms of East, not west. These were really just clean up changes as I happened to be looking at these files

PR #1055, opened 2022-04-16
[sim] Implemented False Positives Display

michmli:
This adds the display of false positives in the rover's fov, but haven't integrated with the target_detector (so rover isn't detecting false positives yet). 
[
Screen Shot 2022-04-16 at 3 47 16 PM
](url)


cgiger00:
what do you mean doesn't have the target detector?

pip_deps/jetson_requirements.txt
```diff
@@ -1,3 +1 @@
 numpy==1.16.2
```
cgiger00, line 1: add these dependencies back?

cgiger00 (changes requested):

michmli:
> 
So far, we've only added the functionality of displaying false positives to the sim, but the rover doesn't detect them as false positive tags.

PR #1056, opened 2022-04-16
[teleop] Handle request to calibrate Joint B

bli76:
- Sends message to move joint B when false message received from /joint_b_calibration_data
- Displays message on GUI while joint B isn't calibrated
- Callback functions in jetson/teleop now check for open-loop state

image


base_station/gui/src/components/ArmControls.vue
```diff
@@ -6,6 +6,9 @@
     <div class="keyboard">
       <GimbalControls/>
     </div>
+    <div v-if='!this.joint_b_is_calibrated'>
+      Calibrating Joint B...
```
CameronTressler, line 10: Can you make this red/bolded/something else more obvious?

CameronTressler: (For all relevant GUIs)

CameronTressler (approved):

tabiosg (commented):
So what exactly is the logic just to be sure? If the user requests to calibrate joint B, are they constantly sending the message to calibrate joint B (They should or else esw's end won't work). Also, how does the user break out of the calibration sequence (can they put in manual commands)?

CameronTressler:
> > So what exactly is the logic just to be sure? If the user requests to calibrate joint B, are they constantly sending the message to calibrate joint B (They should or else esw's end won't work). Also, how does the user break out of the calibration sequence (can they put in manual commands)?
> 
> We have a callback function subscribing to /joint_b_calibration_data, and when we receive a false message, we stop everything and send a single command simulating turning the joystick to max throttle for joint B.
> 
> ig it's not a good idea to immediately start calibrating when the arm becomes uncalibrated, could we use a pop-up alert asking the user to start the calibration process? We also don't have a way to break out of calibration, ig while it's calibrating, include a button that kills the throttle?

Yeah Idk why I didn't think this through but there should definitely be an abort button to stop calibrating. @guthrie I thought you only wanted one message to drive joint b upwards?

cgiger00:
could we add like a timeout? if it doesn't calibrate after 20 seconds automatically abort?

tabiosg:
At first I only wanted one message but then I realized that the nucleos reset if they don't get constantly get messages. Would it be difficult for teleop to implement that?

CameronTressler:
Yeah so we need to send constant messages and allow for user intervention. A timeout would also be good. A further point of clarification, are you (as the arm operator) planning to hit a button to start calibrating the arm? I thought nucleo_bridge was going to know when it needed to be calibrated but maybe I was misunderstanding

tabiosg:
So nucleo_bridge won't move the arm until it's been calibrated, but the operator will still need to a) either hit the button on the gui or b) just calibrate it themselves by moving joint b up.

tabiosg:
If you don't want to implement the button, then you can just have a big red text box saying "Not calibrated!" which will explain to the operator why joint b isn't moving

cgiger00:
can you guys manage the conflicts in this?

CameronTressler:
I think @bli76 has more commits to push as well, some progress was made at the beginning of today

bli76:
Updated logic: when joint B is uncalibrated, a button will appear on the GUI to calibrate the arm. Pressing this button changes the control state to "calibrating," which IK will interpret as off. From here, jetson/teleop will send an open loop command for each false message received from /joint_b_calibration_data, until calibration is either finished or aborted. Calibration will automatically abort after 20s.

Currently, there's nothing blocking open or closed loop from switching back on and aborting the calibration, but ESW should prevent control of the arm while uncalibrated and calibration can continue from where it left off.

image
image

base_station/gui/src/components/ERDTask.vue
```diff
@@ -150,12 +150,13 @@ export default {
...
         {'topic': '/temperature', 'type': 'Temperature'},
         {'topic': '/ra_offset_pos', 'type': 'RAPosition'},
-        {'topic': '/arm_control_state_to_gui', 'type': 'ArmControlState'},
+        {'topic': '/arm_control_state', 'type': 'ArmControlState'},
         {'topic': '/debugMessage', 'type': 'DebugMessage'},
         {'topic': '/drive_vel_data', 'type': 'DriveVelData'},
         {'topic': '/drive_state_data', 'type': 'DriveStateData'},
         {'topic': '/ik_reset', 'type': 'Signal'},
-        {'topic': '/wrist_turn_count', 'type': 'WristTurnCount'}
+        {'topic': '/wrist_turn_count', 'type': 'WristTurnCount'},
```
cgiger00, line 158: see above

jetson/teleop/src/__main__.py
```diff
@@ -210,6 +217,26 @@ def send_sa_kill(self):
         foot_msg.microscope_triad = 0
         lcm_.publish('/foot_openloop_cmd', foot_msg.encode())
 
+    def joint_b_calibration_callback(self, channel, msg):
+        if JointBCalibration.decode(msg).calibrated or self.arm_control_state != 'calibrating':
+            return
+
+        if self.arm_type == self.ArmType.RA:
+            raMotorsData = [0, 1, 0, 0, 0, 0]
```
cgiger00, line 225: have you guys verified 1 goes up instead of -1?

cgiger00: also is there a particular reason teleop is handling the calibration sequence instead of esw?

cgiger00:
plz also resolve the conflicts!

cgiger00:
> Updated logic: when joint B is uncalibrated, a button will appear on the GUI to calibrate the arm. Pressing this button changes the control state to "calibrating," which IK will interpret as off. From here, jetson/teleop will send an open loop command for each false message received from /joint_b_calibration_data, until calibration is either finished or aborted. Calibration will automatically abort after 20s.
> 
> Currently, there's nothing blocking open or closed loop from switching back on and aborting the calibration, but ESW should prevent control of the arm while uncalibrated and calibration can continue from where it left off.
> 
> image image

wym by esw should prevent control of the arm during this?

jetson/teleop/src/__main__.py
tabiosg: Jetson teleop 1 is up, yes

bli76:
> wym by esw should prevent control of the arm during this?

this:

> So nucleo_bridge won't move the arm until it's been calibrated, but the operator will still need to a) either hit the button on the gui or b) just calibrate it themselves by moving joint b up.

Although I think this only applies to closed-loop so ignore what I said about open-loop. @tabiosg can you clarify?

bli76:
Are we still PR'ing into travis-free/main or another branch?

tabiosg:
kinda late on the review but here's some clarification:
1) esw won't prevent the user from moving the joints if it's not calibrated. i think that this should stay true. You guys just need to be cognizant of this info before starting IK
2) esw currently posts data on calibration status every 1 second (1000 ms)

what this pr does well:
1) the user is notified on the gui if joint b is not calibrated
2) if the user wants to calibrate joint b, all it does is send open loop commands until it is calibrated.

what this pr should do before i accept:
1) you should be able to abort. in your comment you say "Currently, there's nothing blocking open or closed loop from switching back on and aborting the calibration" but it seems like there's an abort button here:

<div class="joint-b-calibration" v-if='!this.jointBIsCalibrated && this.controlMode === "calibrating"'>
      Calibrating Joint B...
      <button v-on:click="abortCalibration()">Abort</button>
    </div>
2) idk how easy it is to implement, but currently in order to abort closed loop, you can just move the joystick. can we do something similar here or is that complicated?
 
Also yes, I think we can still continue pushing to travis-free

tabiosg (changes requested):
See msg posted earlier

tabiosg:
Basically after looking at your changes, it would just nice to have a way of aborting calibration

bli76:
The abort button works as intended, and it will also automatically abort after 20 seconds. I'm saying that open/closed loop control isn't blocked during calibration. If someone tries to run IK or use the joystick during calibration, I'm pretty sure that will also abort calibration and we will end up back at square 1 with the uncalibrated arm, which should be fine, they can just start calibration up again.

> esw currently posts data on calibration status every 1 second (1000 ms)

Should I then increase the rate that the simulated open loop commands are being sent? Like every 20ms or something instead of every 1sec? Or is that unnecessary?

I'll also fix the merge conflicts and remove the wrist turn count stuff, all of this will probably happen after I'm done with finals.

tabiosg:
Ok honestly this sounds pretty good the way it is. Idk if Cameron wants to do anything about blocking closed loop control but it sounds good to me tbh

tabiosg:
Just rerequest me to review whenever you're ready

tabiosg:
And I don't think you need to spam open loop commands every 5 ms, every second should be enough for now

bli76:
Should the base branch be ik-integration or main?

CameronTressler:
> Should the base branch be ik-integration or main?

ik-integration. Both ik-integration and ish-integration were merged with main (via travis-free). Now we've recreated both ik and ish integration so we can continue to push changes that need to be tested there

CameronTressler (approved):
Looks good to me

tabiosg (approved):

tabiosg (approved):

CameronTressler:
Closing for now since we should be going into travis-free

PR #1058, opened 2022-04-16
[teleop] toggle button component, heater status indicator 

ErikLytle:
- Made toggle button into a simplified component (different from previous PR)
- Changed Amino Heater & Auto Shutdown to have a toggle for the intended value, with a status light for actual heater status (enabled)

image


base_station/gui/src/components/ToggleButton.vue
```diff
@@ -0,0 +1,118 @@
...
+        <input type="checkbox" :id="id + '_button'" v-model="checkedValue">
+        <span class="toggle__switch"></span>
+    </label>
+</template>
+
+<script>
+export default {
+    data() {
+        return {
+            currentState: this.currentState
```
CameronTressler, line 15: Seems weird cause this would make it so this.currentState is a bit ambiguous. Is it pointing to the member of data() or props? I think this can be removed altogether though I'm not 100$ sure

base_station/gui/src/components/ToggleButton.vue
```diff
@@ -0,0 +1,118 @@
...
+        }
+    },
+
+    computed: {
+        checkedValue: {
+            get() {
+                return this.currentState;
+            },
+            set(newValue) {
+                this.currentState = newValue;
```
CameronTressler, line 44: Not sure this is necessary. It's throwing a bunch of console errors because you're not supposed to set the value of a prop. I think it should be fine to just emit the change, and the parent of this component will update the currentState, which should carry down to this component as a prop

cgiger00:
what should the base branch be changed to?

CameronTressler:
I think we should close this. Opened #1076 instead

CameronTressler:
Didn't mean to add myself. Changes were made by Erik

CameronTressler (approved):

tabiosg (approved):

PR #1062, opened 2022-04-16
[teleop] Fix Chart JS dependency in GUI

CameronTressler:


CameronTressler:
Now builds without warnings

cgiger00 (approved):

amszuch (approved):
Quick note: we only went to this version because of the way the PID gui is already written. There are newer versions of chart.js and vue-chartjs that might make more sense if we decide to make large changes to the gui

cgiger00:
sounds good

PR #1063, opened 2022-04-16
[teleop] Handle error when invalid preset is requested

bli76:


cgiger00 (dismissed):

jetson/ra_kinematics/mrover_arm.cpp
```diff
@@ -697,6 +697,11 @@ void StandardArm::lock_joints_callback(std::string channel, LockJoints msg) {
 
 void StandardArm::arm_preset_callback(std::string channel, ArmPreset msg) {
     std::vector<double> angles = arm_state.get_preset_position(msg.preset);
+
+    // invalid preset
+    if (angles.size() == 0) {
```
CameronTressler, line 702: Did you test this? I think get_preset_position might throw an exception if we give it something invalid...

bli76: I did not test it and turns out my changes don't even compile lol

But this part is fine, the getter returns an empty vector if we give it an invalid preset.

CameronTressler: Oh that's kinda wack, the online documentation for at() says it would throw an error (though that's not how I previously remembered it working).

CameronTressler: Nvm I'm dumb, you had to change it to avoid throwing that exception. Looks good to me then

CameronTressler (approved):

jetson/ra_kinematics/mrover_arm.cpp
bli76: It does throw an error, but I got rid of .at()

CameronTressler: yep yep

cgiger00 (approved):

PR #1071, opened 2022-04-19
[loc] IMU bearing correction

rbridges12:
Bearing correction using GPS bearing when driving in a straight line, gets published to `/corrected_bearing` in a `CorrectedBearing` struct every time odometry is published. Please add commits to display corrected bearing instead of target bearing on the GUI arrow (corrected bearing value is `CorrectedBearing.corrected_bearing`). also updated a few things in the IMU readme


jetson/filter/src/__main__.py
```diff
@@ -164,6 +184,21 @@ def _driveVelDataCallback(self, channel, msg):
...
+        self.drive_v_left = drive_vel_cmd.left
+        self.drive_v_right = drive_vel_cmd.right
+
+    def _targetBearingCallback(self, channel, msg):
+        targetBearing = TargetBearing.decode(msg)
+        self.prev_target_bearing = self.target_bearing
+        self.target_bearing = targetBearing.target_bearing
+
+        target_diff = self.target_bearing - self.prev_target_bearing
+        self.target_bearing_rate = target_diff / \
```
niwhsa9, line 198: escape char?

jetson/filter/src/__main__.py
```diff
@@ -136,9 +141,24 @@ def __init__(self):
...
+        self.lcm.subscribe("/target_bearing", self._targetBearingCallback)
+        self.lcm.subscribe("/drive_vel_cmd", self._driveVelCmdCallback)
+
+        # Initializations for magnetometer bearing correction
+        self.last_bearing_correction = time.time()
+        self.bearing_offset = 0
+
+        # TODO: are these inits going to cause incorrect behavior?
+        self.target_bearing = 0
+        self.prev_target_bearing = 0
```
niwhsa9, line 153: this is troublesome. Not sure if u can just set this to 0.

jetson/filter/src/__main__.py
```diff
@@ -164,6 +184,21 @@ def _driveVelDataCallback(self, channel, msg):
...
+
+    def _targetBearingCallback(self, channel, msg):
+        targetBearing = TargetBearing.decode(msg)
+        self.prev_target_bearing = self.target_bearing
+        self.target_bearing = targetBearing.target_bearing
+
+        target_diff = self.target_bearing - self.prev_target_bearing
+        self.target_bearing_rate = target_diff / \
+            (time.time() - self.last_target_bearing_fresh)
+        self.last_target_bearing_fresh = time.time()
```
niwhsa9, line 200: im worried about the state keeping being part of loc code instead of nav telling you the last requested bearing plus last requested time.

jetson/filter/src/__main__.py
```diff
@@ -283,16 +321,48 @@ def _runLKF(self):
...
+                        print("driving straight")
+
+                        if abs(self.target_bearing - imu_bearing) < self.config["Bearing_error_threshold_deg"]:
+                            print("bearing error is low")
+
+                            # if abs(self.target_bearing_rate) < self.config["Bearing_rate_of_change_threshold_dps"]:
+                            #     print("bearing rate of change is high")
+
+                            old_bearing = imu_bearing + self.bearing_offset
+                            self.bearing_offset = gps_bearing - \
```
niwhsa9, line 347: imu bearing and gps bearing are same frame and same limit?? Are you sure you can do this?

jetson/filter/src/__main__.py
```diff
@@ -283,16 +321,48 @@ def _runLKF(self):
...
+            if gps_bearing != 999:
+                print("valid gps bearing")
+
+                if self.nav_state == "Drive":
+                    print("drive state")
+
+                    if abs(self.drive_v_left - self.drive_v_right) < self.config["Drive_vel_diff_percent"]:
+                        print("driving straight")
+
+                        if abs(self.target_bearing - imu_bearing) < self.config["Bearing_error_threshold_deg"]:
```
niwhsa9, line 340: don't you want this to go the other way, when the error is too high correct?

niwhsa9 (changes requested):

jetson/filter/src/__main__.py
rbridges12: thats just from the python autoformatter but I'll get rid of it

niwhsa9: oh well ok thats fine then, if linter is happy im happy.

jetson/filter/src/__main__.py
rbridges12: I actually think this should be fine. Since we have to be driving in order to correct the bearing, and I assume we will always have a target bearing if we are driving somewhere, then these initial values should never actually be used to check the correction condition. ig its possible that the very first iteration of nav could cause trouble, like if weve only received a single target bearing so the previous is still the default zero, but even then I doubt that all the other conditions would pass on the very first iteration of nav driving. We'll find out for sure in testing ig. Also FYI this is currently not even using the previous target bearing or rate of target bearing, we're gonna test it without that condition check first.

jetson/filter/src/__main__.py
rbridges12: yeah ig there might be a problem with time inaccuracy due to network latency and LCM packet loss. again, I'm not sure if we're even gonna need these metrics for the correction, do you think its worth adding this to nav before we even test it?

jetson/filter/src/__main__.py
rbridges12: yeah this is definitely wrong because I'm never modding by 360, let me think about how to fix this

jetson/filter/src/__main__.py
rbridges12: no because this isnt error between actual bearing and measured bearing, this is error between target bearing and measured bearing. So if we're driving straight, nav will always line up the rover so that the error between target and measured is close to zero, thats what our pid loop does. This is just another check to make sure we're driving straight and not turning, ig its kinda redundant now that I've added the drive velocity command check. The thing I *did* get backwards was the check on rate of target bearing change which I will fix

jetson/filter/src/__main__.py
niwhsa9: ok, as long as we are sure that this won't get triggered on the first iter thats fine, i guess we'd just add a flag for first iteration if it did. Like you said - should be obvious.

jetson/filter/src/__main__.py
niwhsa9: oh right, my bad.

jetson/filter/src/__main__.py
qhdwight: Usually prefer to use `None` in Python when a variable has "no data" and then handle it appropriately later with if statements

rbridges12: yeah I'll do that, I was being lazy

config/filter/config.json
```diff
@@ -21,5 +21,11 @@
     "IMU_accel_threshold": 0.5,
 
     "RefCoords": {"lat": 42.2766, "long": -83.7382},
```
cgiger00, line 23: are these the approx coordinates of ann arbor rn?

jetson/filter/src/__main__.py
```diff
@@ -164,6 +184,21 @@ def _driveVelDataCallback(self, channel, msg):
         drive_vel_data = DriveVelData.decode(msg)
         self.encoder_velocities[drive_vel_data.axis] = drive_vel_data.vel_percent
```
cgiger00, line 185: ig this might break soon, i think i just merged a pr that makes it vel_m_s instead of vel_percent

jetson/filter/src/__main__.py
```diff
@@ -164,6 +184,21 @@ def _driveVelDataCallback(self, channel, msg):
...
 
+    def _driveVelCmdCallback(self, channel, msg):
+        drive_vel_cmd = DriveVelCmd.decode(msg)
+        self.drive_v_left = drive_vel_cmd.left
+        self.drive_v_right = drive_vel_cmd.right
+
+    def _targetBearingCallback(self, channel, msg):
+        targetBearing = TargetBearing.decode(msg)
+        self.prev_target_bearing = self.target_bearing
```
cgiger00, line 194: readability: self.current_target_bearing instead?

jetson/filter/src/__main__.py
cgiger00: i think it'll be easier to assess after we do the long time test saturday so probably fine not to add it in before, your call tho

jetson/filter/src/__main__.py
```diff
@@ -120,7 +124,8 @@ def __init__(self):
             self.config = json.load(config)
 
         self.gps = Gps()
-        self.imu = Imu(self.config["IMU_accel_filter_bias"], self.config["IMU_accel_threshold"])
+        self.imu = Imu(self.config["IMU_accel_filter_bias"],
+                       self.config["IMU_accel_threshold"])
         self.nav_state = None
         self.static_nav_states = {"Off", "Done", "Search Spin Wait", "Turned to Target Wait", "Gate Spin Wait",
```
cgiger00, line 130: these are no longer correct right?

jetson/filter/src/__main__.py
```diff
@@ -227,10 +262,12 @@ def _runLKF(self):
         vel = None
         accel = self._getFreshAccel(bearing)
 
-        u = np.array([accel["north"], accel["east"]]) if accel is not None else None
+        u = np.array([accel["north"], accel["east"]]
```
cgiger00, line 265: is it backwards on the gui that the gps coordinates are north & west?

jetson/filter/src/__main__.py
```diff
@@ -283,16 +321,48 @@ def _runLKF(self):
...
+        gps_bearing = self.gps.bearing.bearing_deg
+        imu_bearing = self.imu.bearing.bearing_deg
+
+        if time.time() - self.last_bearing_correction > self.config["Bearing_offset_interval_sec"]:
+            print("timeout passed")
+
+            if gps_bearing != 999:
+                print("valid gps bearing")
+
+                if self.nav_state == "Drive":
```
cgiger00, line 334: no longer exists?

jetson/filter/src/__main__.py
```diff
@@ -283,16 +321,48 @@ def _runLKF(self):
...
         Returns a fresh bearing to use. Uses IMU over GPS, returns None if no fresh sensors
 
         @return float/None: bearing (decimal degrees East of North)
         '''
+
+        self.correct_bearing()
+
         if time.time() - self.imu.last_fresh <= self.config["IMU_fresh_timeout"]:
-            return self.imu.bearing.bearing_deg
+            return self.imu.bearing.bearing_deg  # + self.bearing_offset
```
cgiger00, line 363: comment as a (maybe) todo to add back in?

jetson/imu_channel/README.md
```diff
@@ -58,17 +58,49 @@ You should see a stream of `IMUData` structs being printed to the terminal, try
...
-Now just detach the BNO055 from its mount and wave it around in the air. Watch [this video](https://www.youtube.com/watch?v=Bw0WuAyGsnY) on what patterns to move it in for optimal results. You'll know you're done when all three calibration status variables read 3.
-
-_(There's also a 4th calibration status called sys, but that one usually sorts itself out. I'll add it to the LCM struct in the future)_
+Be aware that the accelerometer can be very tricky to calibrate, sometimes taking upwards of 10 minutes to get the calibration status to 3. For fastest calibration, follow the pattern in the video *very* slowly, rotating between angles slowly and holding the IMU still at those angles for around 10 seconds each.  It's also worth noting that the magnetometer calibration status often drops to a lower value (we've seen anywhere from 2 to 0) after running it for a while.
 
 The BNO055 should now be ready to go and outputting good data, so you can run whatever other programs need to use it.
 
+## Arduino Pro CLI
+
+Instead of using the Arduino IDE to upload to the Arduino, you can use the Arduino Pro CLI tool. This is better since it can be done over SSH purely from the command line.
```
cgiger00, line 71: nice

jetson/imu_channel/README.md
```diff
@@ -58,17 +58,49 @@ You should see a stream of `IMUData` structs being printed to the terminal, try
...
+Exit screen by pressing _ctrl+a_ then _k_
+
+
 ## TODO
-- update `IMUData` LCM struct to include all 4 calibration status values and maybe get rid of values we don't need?
 - add feature to allow axes to be rearranged easily (to account for different sensor orientation relative to rover)
-- try it out on Arduino IDE 2.0 and change tutorial to work with that, or try it on Arduino Pro CMD and write a script that does everything using that
\ No newline at end of file
+- try it out on Arduino IDE 2.0 and change tutorial to work with that
+- write a script that uses arduino cli to do everything automatically
```
cgiger00, line 106: 'add in bearing correction if necessary' or smth of the sort

cgiger00, line 106: would be very useful! could be modified to work for the spectral triad too ☠️

jetson/filter/src/__main__.py
```diff
@@ -404,6 +476,13 @@ async def run(self):
 
                 odom = self.state_estimate.asOdom()
                 self.lcm.publish('/odometry', odom.encode())
+
+                corrected_bearing = bearing + self.bearing_offset
+                corrected_bearing_struct = CorrectedBearing()
```
cgiger00, line 481: i guess to save some space you could use the same Odometry object but publish the corrected_bearing to this one's bearing and put it to "/corrected_bearing", gives extra info but i think for testing it's an easy add on so we don't need to delete the CorrectedBearing lcm after we're done testing/leave a random unused one hanging

cgiger00: would put extra unnecessary overhead onto the network tho but i dont think it's a big deal for testing

jetson/filter/src/__main__.py
qhdwight: Here are the correct status names:

```c++
// Gets the string representation of a nav state.
std::string StateMachine::stringifyNavState() const {
    static const std::unordered_map<NavState, std::string> navStateNames =
            {
                    {NavState::Off,                  "Off"},
                    {NavState::Done,                 "Done"},
                    {NavState::DriveWaypoints,       "Drive Waypoints"},
                    {NavState::BeginSearch,          "Change Search Algorithm"},
                    {NavState::Search,               "Search"},
                    {NavState::DriveToTarget,        "Drive to Target"},
                    {NavState::TurnAroundObs,        "Turn Around Obstacle"},
                    {NavState::DriveAroundObs,       "Drive Around Obstacle"},
                    {NavState::SearchTurnAroundObs,  "Search Turn Around Obstacle"},
                    {NavState::SearchDriveAroundObs, "Search Drive Around Obstacle"},
                    {NavState::BeginGateSearch,      "Gate Prepare"},
                    {NavState::GateTraverse,         "Gate Traverse"},

                    {NavState::Unknown,              "Unknown"}
            };

    return navStateNames.at(mRover->currentState());
} // stringifyNavState()
```

qhdwight: I'm now noticing `BeginSearch` is named wrong

config/filter/config.json
rbridges12: yep, ig a sidenote is we need to remember to change this for URC and CIRC

jetson/filter/src/__main__.py
rbridges12: yeah ill merge that into my branch before this gets merged

jetson/filter/src/__main__.py
rbridges12: that list of static nav states is left over from the KF, it isn't actually used at all rn. on that note, should I just remove all the KF code since we aren't gonna be using it?

jetson/imu_channel/README.md
rbridges12: yeah thats exactly what I was thinking, should be like 2 lines long if arduino CLI and all the necessary libs are already installed on the base station

jetson/filter/src/__main__.py
rbridges12: I mean either way we still have to remember to delete something after testing it, I think its gonna be easier to just leave it like this and ill be sure to get rid of that LCM struct once were done testing this

config/filter/config.json
cgiger00: what should they be changed to for each respective comp?

jetson/filter/src/__main__.py
cgiger00: honestly, yeah, if you chose to rewrite it you can always use the git history as a place to start, or start w smth new

config/filter/config.json
rbridges12: I think any GPS point within a few miles of the comp should be fine

jetson/filter/src/__main__.py
niwhsa9: current is redundant

jetson/filter/src/inputs.py
```diff
@@ -252,7 +252,7 @@ def update(self, new_bearing_sensor):
         if hasattr(new_bearing_sensor, "bearing"):
             self.bearing_deg = new_bearing_sensor.bearing
         elif hasattr(new_bearing_sensor, "bearing_deg"):
-            if new_bearing_sensor.bearing_deg != 999.:
+            if new_bearing_sensor.bearing_deg != 999:
```
cgiger00, line 255: can this be a constant or commented somewhere as the invalid value?

PR #962, opened 2022-03-17
[teleop, nav, percep, loc, sim] auton integration

cgiger00:

PR #1072, opened 2022-04-19
[percep -> nav] Use point cloud to better retrieve AR tag distance and bearing

qhdwight:
Previously we were using just the pixel information + camera FOV to calculate bearing. Distance was calculated using the depth, so it was actually inaccurate and would be worse the less aligned you were with the gate. Now the actual point cloud information from the ZED is being used to calculate both of these better.

The intent of these fixes were to improve the accuracy of nav's estimation of gate positions since it uses both of these values. They should still be filtered on nav's end.

### DIFF
--- a/jetson/percep/main.cpp
+++ b/jetson/percep/main.cpp
@@ -7,40 +7,40 @@
 using namespace cv;
 using namespace std;
 using namespace std::chrono_literals;
- 
+
 int main() {
 
- /* --- Reading in Config File --- */
-  rapidjson::Document mRoverConfig;
-  ifstream configFile;
-  string configPath = getenv("MROVER_CONFIG");
-  configPath += "/config_percep/config.json";
-  configFile.open( configPath );
-  string config = "";
-  string setting;
-  while( configFile >> setting ) {
-    config += setting;
-  }
-  configFile.close();
-  mRoverConfig.Parse( config.c_str() );
+    /* --- Reading in Config File --- */
+    rapidjson::Document mRoverConfig;
+    ifstream configFile;
+    string configPath = getenv("MROVER_CONFIG");
+    configPath += "/config_percep/config.json";
+    configFile.open(configPath);
+    string config = "";
+    string setting;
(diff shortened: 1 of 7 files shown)


qhdwight:
As per usual view with whitespace off: https://github.com/umrover/mrover-workspace/pull/1072/files?diff=split&w=1

niwhsa9 (approved):
[https://media.giphy.com/media/gLu90OMjz4j3hAkJk2/giphy.gif](url)

qhdwight:
```c++
int y_pixel = tags.y_pixels.at(i);
int x_pixel = tags.x_pixels.at(i);
float x, y, z;
{
    auto xyz = xyz_img.at<Vec4f>(y_pixel, x_pixel);
    x = xyz[0];
    y = xyz[1];
    z = xyz[2];
}
auto raw_depth = depth_img.at<float>(y_pixel, x_pixel);
if (!isnan(raw_depth)) {
    arTags[i].distance = sqrt(x * x + y * y + z * z) * static_cast<float>(MM_PER_M);
}
// +z is forward, +x is right, all relative to camera
arTags[i].bearing = atan2(x, z);
arTags[i].id = tags.ids.at(i);
tags.buf_counts[i] = 0;
```

this is the important part of this pr

jetson/percep/main.cpp
```diff
@@ -7,59 +7,59 @@
 using namespace cv;
 using namespace std;
```
cgiger00, line 8: would you mind getting rid of this? bad practice

qhdwight: yeah it was in the original code but I can remove the `using namespace`s

cgiger00: thx

cgiger00:
do you want to hold on off merging this before this PR https://github.com/umrover/mrover-workspace/pull/962 is approved?

qhdwight:
> do you want to hold on off merging this before this PR #962 is approved?

Yes that would probably be good

PR #1075, opened 2022-04-24
[nav] Lower chance gate will be regenerated incorrectly

qhdwight:
While the rover was going to the prep point it saw the target and regenerated an incorrect gate position. Try to mitigate this by:

1) Increasing the filter size from 4 to 8
2) Not allowing gate path regeneration while going to the prep point


ankithu (approved):
lgtm

PR #1083, opened 2022-04-28
[percep] fix failing compilation bc of missing std::

cgiger00:


qhdwight (approved):
Look's good, all things prefixed with std:: don't seem to have possible clashes with other symbols

niwhsa9 (approved):
ok

PR #1092, opened 2022-05-05
[sim] False pos in 2d sim

danieljiang520:
https://user-images.githubusercontent.com/71047773/166873821-25f28cbf-c532-4484-9765-22e123c46a5e.mov


simulators/nav/runsim
```diff
@@ -28,7 +28,7 @@ build_jarvis() {
     if $nav_flag && $lcm_flag  && $sim_flag;
     then
         printf "${CYAN}Building nav + lcm + sim\n${NC}"
-        vagrant ssh -c './jarvis build jetson/nav && ./jarvis build lcm_bridge/server && ./jarvis build simulators/nav && (jarvis exec jetson/nav & jarvis exec lcm_bridge/server & jarvis exec simulators/nav); /bin/bash'
+        vagrant ssh -c './jarvis build jetson/nav && ./jarvis build lcm_bridge/server && ./jarvis build simulators/nav && (./jarvis exec jetson/nav & ./jarvis exec lcm_bridge/server & ./jarvis exec simulators/nav); /bin/bash'
```
cgiger00, line 31: interesting

cgiger00:
is it possible to add false positives of different IDs like the range of the aruco lib is in the hundreds

cgiger00 (commented):
see comment above

danieljiang520:
> is it possible to add false positives of different IDs like the range of the aruco lib is in the hundreds

Right now, all of the false positives are like clones of the real ar tags on the field, and the number beside the red triangle is the id of the real ar tags that it cloned from. It is possible to have multiple false positives with the same reference to the real ar tag (i.r. same id), or with different references to the different real ar tags. So, if there are 100 of real ar tags on the field, it is possible to have false positives of different IDs in the range of 0-99 or whatever the ids of the real ar tags are.
Screen Shot 2022-05-09 at 18 33 41

Currently, I set the maximum number of false positives at one instance to not exceed the number of real ar tags on the field and not exceed the "Max # of false positives" setting. Do you mean eliminating this threshold to make the sim be able to generate hundreds of false positives at one instance?

cgiger00:
> > is it possible to add false positives of different IDs like the range of the aruco lib is in the hundreds
> 
> Right now, all of the false positives are like clones of the real ar tags on the field, and the number beside the red triangle is the id of the real ar tags that it cloned from. It is possible to have multiple false positives with the same reference to the real ar tag (i.r. same id), or with different references to the different real ar tags. So, if there are 100 real ar tags on the field, it is possible to have false positives of different IDs in the range of 0-99 or whatever the ids of the real ar tags are. Screen Shot 2022-05-09 at 18 33 41
> 
> Currently, I set the maximum number of false positives at one instance to not exceed the number of real ar tags on the field and not exceed the "Max # of false positives" setting. Do you mean eliminating this threshold to make the sim be able to generate hundreds of false positives at one instance? @cgiger00

no i meant more like the false positives can range from 0 - 100 but not necessarily be showing hundreds, like it can show 3 false positives say ranging from either 0 - 100

zwgold (approved):

ankithu (approved):

PR #1095, opened 2022-05-20
[nav] intermediate search points

ankithu:
adds intermediate search points every 3 meters. This hasn't been tested in the sim due to difficulty with the new enabled interface, but visually it looks like the generated points are ok:

image


ankithu:
need to fix this pr to only include the changes to search stuff mb.

PR #1096, opened 2022-05-20
[nav] intermediate search points + enter search early + tighter target tolerance

ankithu:
adds intermediate search points every 3 meters. Tested in sim.

image


cgiger00 (commented):
lgtm

cgiger00 (approved):

PR #1102, opened 2022-05-24
[nav] should fix ar tag id issue and gate targeting bad post issues + remove intermediate search points

ankithu:
We were allowing for baseGateID + 1 AR tags reads to go through on AR tag post legs which is not what we want to do so I removed that. We were not actually checking IDs for the gate in the case that we only see one tag. The code is slightly redundant as is and probably can be done better but I think redundancy in this sort of thing is probably fine - tested in sim.


jetson/nav/environment.cpp
```diff
@@ -57,7 +57,7 @@ void Environment::updatePost(std::shared_ptr<Rover> const& rover, std::shared_pt
                 mPostOneLong.push(postOdom.longitude_deg + postOdom.longitude_min / 60.0);
                 mHasNewPostUpdate = true;
             }
-            if (rightTarget.id == mBaseGateId + 1) {
+            if (rightTarget.id == mBaseGateId + 1 && mShouldLookForGate) {
```
cgiger00, line 60: mLookForGate & beyond might be cleaner?

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -40,7 +40,7 @@ NavState SearchStateMachine::executeSearch() {
     std::shared_ptr<Rover> rover = sm->getRover();
 
     Target const& leftTarget = env->getLeftTarget();
-    Target const& rightTarget = env->getLeftTarget();
+    Target const& rightTarget = env->getRightTarget();
```
cgiger00, line 43: oof good catch

cgiger00 (approved):
lgtm, added a style suggestion

PR #1098, opened 2022-05-20
[esw, teleop]  add four camera config to all missions

tabiosg:
esw code itself has been tested (it properly changes to stream to one laptop -- has only been tested with one laptop but i'd assume it would still work with two laptops). still need teleop to implement four cameras being clickable on the gui and have it send the Mission.lcm to /cameras_mission

PR #1103, opened 2022-05-24
[esw] fix cameras crashing

tabiosg:
Added a try/exception when trying to capture and render image. If capture and rendering the image from one camera fails, it will stop streaming to that camera. The program should theoretically not crash. The operator the GUI will not have any way of telling that the program stopped streaming that camera other than by looking at the feed itself (all the other cameras should still be working). Since the program theoretically no longer crashes, the operator should be able to quickly reopen that particular camera.

Camera code is very similar to what's in #1098. This is untested but hopefully it will prevent crashing.

### DIFF
--- a/jetson/cameras/src/__main__.py
+++ b/jetson/cameras/src/__main__.py
@@ -11,11 +11,14 @@ __pipelines = [None] * 4
 ARGUMENTS_LOW = ['--headless', '--bitrate=300000', '--width=256', '--height=144']
 # 10.0.0.1 represents the ip of the main base station laptop
 # 10.0.0.2 represents the ip of the secondary science laptop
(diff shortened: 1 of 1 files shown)


jetson/cameras/src/__main__.py
```diff
@@ -11,20 +11,37 @@
 ARGUMENTS_LOW = ['--headless', '--bitrate=300000', '--width=256', '--height=144']
 # 10.0.0.1 represents the ip of the main base station laptop
 # 10.0.0.2 represents the ip of the secondary science laptop
-remote_ip = ["10.0.0.1:5000", "10.0.0.1:5001", "10.0.0.2:5000", "10.0.0.2:5001"]
+ips_one_laptop = ["10.0.0.1:5000", "10.0.0.1:5001", "10.0.0.1:5002", "10.0.0.1:5003"]
```
cgiger00, line 14: we mentioned the naming

cgiger00 (approved):
seemed to work well today!

PR #1104, opened 2022-05-24
[esw] Add zed gimbal motor code

tabiosg:


tabiosg:
Making this PR since it may be useful as a reference for next year (if the team decides to use brushless motors).

tabiosg:
Ok this fails the build but we don't need to accept this PR any time soon... we can just keep this as a PR for now and refer to it for next year if needed

cgiger00:
sad...

PR #1001, opened 2022-04-02
[percep] opencv AR tag debugging tool 

JasonLiu2002:
Shows thresholded matrix and sliders for tuning ARUCO params

PR #1105, opened 2022-05-25
Fixed ar tag debug tool bugs from PR #1001

JasonLiu2002:


cgiger00 (commented):
were you able to run this? i made all these fixes and it still seg faulted

cgiger00 (approved):
works on the rover

PR #1107, opened 2022-05-25
[admin] symlink rules for imu and gps

cgiger00:
added auton services/modified relevant files (zed one still doesn't quite work)


ansible/roles/jetson_services/tasks/main.yml
```diff
@@ -3,12 +3,13 @@
       src: "files/{{ item }}.service"
       dest: "/etc/systemd/system/{{ item }}.service"
   with_items:
+      - rover-jetson-auton-filter
```
CameronTressler, line 6: nit: keep alpha order?

CameronTressler: Preference thing but not sure why we wouldn't

CameronTressler (approved):

rbridges12 (approved):
very cool, looks good

ansible/roles/jetson_services/tasks/main.yml
cgiger00: fixed

PR #1113, opened 2022-05-25
[nav] fix using wrong search file

ankithu:


zwgold (approved):
merge

zwgold:
Also, should we change the distance threshold for the post to something other than 1.0? I saw at testing the results said it was too short (was 1.5 originally, could we make it a happy medium of 1.25, or on the higher side bc of the ZED Gimbal placement? Or wait to test it in person to decide?)

ankithu:
> Also, should we change the distance threshold for the post to something other than 1.0? I saw at testing the results said it was too short (was 1.5 originally, could we make it a happy medium of 1.25, or on the higher side bc of the ZED Gimbal placement? Or wait to test it in person to decide?)

gonna leave at 1 for now because we're changing the angle of the zed to hopefully make that possible. But will change if not

PR #1115, opened 2022-05-26
[sim] added enable in 2D sim

danieljiang520:
2D sim now subscribes to Enable.lcm in the /auton_enabled channel

**NOTE: MAKE SURE ENABLELCM BUTTON IS OFF WHEN RUNNING THE 2D SIM**
Doing so will stop 2d sim publish everything except for odom and obstacle list


simulators/nav/src/components/NavSimulator.vue
```diff
@@ -440,6 +448,7 @@ export default class NavSimulator extends Vue {
       /* Subscriptions */
       [
         { topic: '/autonomous',     type: 'Joystick' },
+        { topic: '/auton_enabled',  type: 'Enable' },
```
cgiger00, line 451: probably also add the teleop_enabled? @ankithu?

simulators/nav/src/components/NavSimulator.vue
```diff
@@ -355,6 +358,8 @@ export default class NavSimulator extends Vue {
 
   /* Publish the given LCM message on the given channel. */
   private publish(channel:string, payload:any, odomOrObstacle:boolean):void {
+
```
zwgold, line 361: `simulators/nav` won't compile unless you delete this blank line on 361 (linting issue) when I tried to run it

simulators/nav/src/components/NavSimulator.vue
ankithu: not sure it really needs to subscribe to anything

danieljiang520: Just reverted back to only publish enable

ankithu:
Tried running with nav, doesn't seem to start the rover when I click Auton State on. Not sure why enable lcm should be switched off, basically what this PR should do is just publish an Enable struct to /auton_enabled instead of the previous AutonState struct.

ankithu:
Tried running with nav, doesn't seem to start the rover when I click Auton State on. Not sure why enable lcm should be switched off, basically what this PR should do is just publish an Enable struct to /auton_enabled instead of the previous AutonState struct. Maybe I'm testing it wrong tho? Are you guys able to get it to work?

simulators/nav/src/components/NavSimulator.vue
```diff
@@ -454,7 +454,7 @@ export default class NavSimulator extends Vue {
 
     /* Set up publishing LCMs */
     this.intervalLcmPublish = window.setInterval(() => {
-      this.publish('/auton', { type: 'AutonState', is_auton: this.autonOn }, false);
+      this.publish('/auton_enabled', { type: 'Enable', is_auton: this.autonOn }, false);
```
zwgold, line 457: Should be `enabled` and not `is_auton` @ankithu @danieljiang520 

@danieljiang520 along with that, you also want to change the `is_auton` found `simulators/nav/src/.eslintrc.js`, it's located twice in there, and note that the order of the variables is alphabetical which we'll want to keep

danieljiang520:
Don't need to disable enablelcm anymore since sim is just publishing
https://user-images.githubusercontent.com/71047773/170836347-c2818974-f389-4949-9c8f-a884b1ce1703.mov

zwgold (approved):
lgtm

cgiger00 (approved):

ankithu (approved):
tested and lgtm

PR #1121, opened 2022-05-27
[percep] Return planar distance instead of 3D

qhdwight:
Currently we are returning 3D distance, but nav expects 2D distance.
This will make issues worse the further the AR tags are up on the post.
Should improve gate performance make make driving to a single tag more reliable.


cgiger00 (approved):
lgtm

ankithu (approved):

cgiger00:
tested, works as expected

PR #1131, opened 2022-05-29
[teleop] Add conversions for chlorophyll spectral

CameronTressler:


CameronTressler:
Message from Gabe:

Here are the numbers for the conversion from spectral 2 and spectral 1 to spectral 0. It's kind of hard to label these clearly but "x1" and "x2" are the measured values currently being printed out for spectral sensors 1 and 2, and "y" is the converted value for each.

Spectral 1: y1 = (1 / 0.94) * (x1 - 23.28)
Spectral 2: y2 = (1 / 0.9969) * (x2 + 32.8575)

base_station/gui/src/components/Chlorophyll.vue
```diff
@@ -114,6 +114,14 @@ export default {
         'device': id,
         'enable': enabled
       })
+    },
+
+    spectral_1_conversion: function(sensor_reading) {
+      return (1 / 0.94) / (sensor_reading - 23.28)
```
tabiosg, line 120: Will the gui break if we get a divide by 0 problem

tabiosg (commented):
I'll test this later and then I'll approve if it works

base_station/gui/src/components/Chlorophyll.vue
tabiosg: Actually, Gabe said it was multiply, not divide

tabiosg (changes requested):
Change to multiply not divide in equation i think

cgiger00 (approved):

tabiosg (approved):

cgiger00:
NOTE: ***AFTER THE CODE FREEZE*** needed addition for science that we had missed

PR #1136, opened 2022-06-01
[teleop] Use JPG map files

CameronTressler:


cgiger00 (approved):
tested on the base station laptop...still needs the folder change fix

PR #1138, opened 2022-06-01
[teleop] Move gimbal controls

CameronTressler:
Now includes auton GUI. Tested with all GUIs, gimbal controls work


tabiosg (commented):
it might cause issues where ish gui will be able to control the mast

PR #1140, opened 2022-06-02
[nav] testing changes

ankithu:
-hexagon search
-separate PIDs
-going to the wrong AR-tag fix


jetson/nav/environment.cpp
```diff
@@ -29,10 +29,18 @@ void Environment::setTargets(TargetList const& targets) {
         mLeftTargetBearing.push(mTargetLeft.get().bearing);
         mLeftTargetDistance.push(mTargetLeft.get().distance);
     }
+    else{
+        mLeftTargetDistance.reset();
+        mLeftTargetBearing.reest();
```
zwgold, line 34: `reset()`

zwgold:
The spiral reaches 20m at the very last point. Would recommend adding 5 more points for the other corners of the hexagon to hit 20m to guarantee us seeing the entire space. Otherwise pattern is solid

zwgold:
Fixes error from earlier where if we see two AR Tags for the single post (one is correct and one is incorrect) and we lose the correct one, it will auto default to the one it currently sees. (which can be the wrong one) A small edge case, but still good to consider.

jetson/nav/environment.cpp
zwgold: Also space after `else`

zwgold (approved):
lgtm, PID separation for turn and drive looks fine and works in sim

jetson/nav/rover.cpp
```diff
@@ -14,6 +14,11 @@ Rover::Rover(const rapidjson::Document& config, lcm::LCM& lcmObject)
                               config["bearingPid"]["kD"].GetDouble())
                               .withMaxInput(360.0)
                               .withThreshold(mConfig["navThresholds"]["turningBearing"].GetDouble())),
+          mDriveBearingPid(PidLoop(config["driveBearingPid"]["kP"].GetDouble(),
+                                   config["driveBearingPid"]["kI"].GetDouble(),
+                                   config["driveBearingPid"]["kD"].GetDouble())
+                                    .withMaxInput(360.0)
```
qhdwight, line 20: U do not want this. This is when the input can wrap around (such as angles)

jetson/nav/rover.hpp
```diff
@@ -69,6 +69,8 @@ class Rover {
 
     PidLoop& bearingPid();
```
cgiger00, line 70: might be more clear to make this turningBearingPid

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -116,18 +122,15 @@ NavState SearchStateMachine::executeDriveToTarget() {
...
-            if (leftTarget.id == lastWaypoint.id) {
-                distance = leftTarget.distance;
-                bearing = leftTarget.bearing + currentBearing;
-            } else {
-                distance = rightTarget.distance;
-                bearing = rightTarget.bearing + currentBearing;
-            }
-        } else if (leftTarget.id >= 0) {
+        //not gate (AR tag)
+        if (leftTarget.id == lastWaypoint.id) {
```
cgiger00, line 126: so it no longer checks +1?

cgiger00 (commented):
left some comments

PR #1152, opened 2022-07-10
[sim] Added 2dsim feature to ignore auton drive control with a button + updated sim to the new enable interface

danieljiang520:
https://user-images.githubusercontent.com/71047773/178129199-fa585e46-e113-45a0-8d0c-e993342a2793.mov

Added a toggle in the current sim that freezes the rover. When 'simulating auton drive control' is turned off, the sim ignores values from the drive LCM and sets the rover velocity to 0.

This old [PR](https://github.com/umrover/mrover-workspace/pull/1115) that updated sim to the new enable interface probably got lost while merging into main. I have added the same changes in this PR

### DIFF
--- a/simulators/nav/src/components/HotKeys.vue
+++ b/simulators/nav/src/components/HotKeys.vue
@@ -48,6 +48,9 @@ export default class HotKeys extends Vue {
   private readonly simulatePercep!:boolean;
 
   @Getter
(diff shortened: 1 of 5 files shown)


ankithu (approved):
changes look good. tested and works. Thanks Daniel!

zwgold (approved):

PR #1155, opened 2022-07-27
[nav] Add Recovery Functionality to Nav 

John-Trager:
Created Recovery object that can be used to detect if the rover is stuck (based on rover displacement over time) and then performa a J turn maneuver to recover.
DONE:
- recovery works in DriveWaypoints state (in sim)
- added condition for recovery when rover is turning vs driving (time based for turning, displacement based for driving)
- added recovery to search state-machine and drive to target state (works in sim)
- fixed drivetotarget issue where recovery would lose the target (will now go back to target after recovering)
- refactored drive function

TODO:
- (issue) slight issue when in search and recovery is performed, part of the search path can be skipped
- (testing) test robustness of recovery detection and maneuvers in sim and IRL. make decision if will go into CIRC code or if too risky
- (optional) use circular buffer vector instead of deque, will perform better but current implementation performance is nominal


jetson/nav/recovery.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+}//Recovery()
+
+bool Recovery::isStuck(){
+    // check displacement (TODO: have condition for if we are turning! or just don't call if we are turning?)
+    // if displacement < threshold: return true
+    // else return false
+
+    double distance = estimateDistance(odoms.front().odom,odoms.back().odom);
+    if (!odoms.empty() && getDuration(odoms.front().time,odoms.back().time) >= offsetAmount && distance < dispThresh){
```
qhdwight, line 15: just do `return <expr>` instead of having if and else

jetson/nav/recovery.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+}//update()
+
+
+int Recovery::getDuration(time_point p1, time_point p2){
+    // get duration between two points in seconds (rounded to nearest second)
+    // let p1 be the newer of the time points and p2 be the later of the time points
+    // for our use case
+    auto milliseconds = std::chrono::duration_cast<std::chrono::milliseconds>(p1 - p2);
+    return int((milliseconds.count()/1000.0) + 0.5);
```
qhdwight, line 48: you can duration_cast to seconds instead of milliseconds and use `count()`. Also why is this /2 ?

jetson/nav/stateMachine.hpp
```diff
@@ -87,4 +88,10 @@ class StateMachine : public std::enable_shared_from_this<StateMachine> {
     std::shared_ptr<ObstacleAvoidanceStateMachine> mObstacleAvoidanceStateMachine;
 
     time_point mTimePoint, mPrevTimePoint;
+
+    // Recovery object
+    // TODO put in config file (change to 15 seconds for compition)
+    int seconds = 8;
```
qhdwight, line 94: use std chrono duration

jetson/nav/recovery.cpp
John-Trager: Unsure what you mean by /2 but this should be resolved now

qhdwight:
Please also format with the corresponding `.clang-format` file in the `nav` directory :)

jetson/nav/recovery.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+/***
+ * Recovery Watchdog
+ * Class used to detect if Rover has become stuck
+ * and generate recovery manuevers to get unstuck.
+ */
+class Recovery {
+public:
+    Recovery(std::chrono::duration<int> offsetAmount, double dispThresh,
+             std::chrono::duration<int> maxTurnTime);
```
qhdwight, line 36: Add a `using` alias for `std::chrono::duration<int>` since that is a mouth full

https://www.nextptr.com/tutorial/ta1193988140/how-cplusplus-using-or-aliasdeclaration-is-better-than-typedef

qhdwight: Can use search and replace in editor

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -53,14 +56,51 @@ NavState SearchStateMachine::executeSearch() {
...
         if (isWantedTarget) {
             return NavState::DriveToTarget;
         }
     }
 
-    Odometry const& nextSearchPoint = mSearchPoints.front();
     double dt = sm->getDtSeconds();
+
+    // Recovery watch for search
+    if (sm->getRecovery()->getRecoveryState()) {
```
qhdwight, line 68: Instead of typing `sm->getRecovery()` a billion times consider making a variable. Same idea with path and what not. This function is a little hard to read.

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -126,13 +166,48 @@ NavState SearchStateMachine::executeDriveToTarget() {
...
-        }
-        else {
+        } else {
             std::cerr << "Lost target" << std::endl;
             return NavState::Search;
         }
     }
 
+    // Recovery watch for DriveToTarget
+    if (sm->getRecovery()->getRecoveryState()) {
```
qhdwight, line 176: Logic seems duplicated? Please try to refactor out common code.

jetson/nav/stateMachine.hpp
```diff
@@ -87,4 +90,12 @@ class StateMachine : public std::enable_shared_from_this<StateMachine> {
...
 
     time_point mTimePoint, mPrevTimePoint;
+
+    // Recovery object
+    // TODO put in config file
+    // Test what this value should be in IRL testing
+    std::chrono::duration<int> secondsOffset = 5s;
+    double meters_threshold = 1;
+    std::chrono::duration<int> maxTurnTime = 15s;
+    std::shared_ptr<Recovery> mRecovery = std::make_shared<Recovery>(secondsOffset,meters_threshold,maxTurnTime);
```
niwhsa9, line 100: it doesn't seem to me like the recovery object needs multiple owners. It's owner is the state machine and its life time should be completely tied to the state machine too. Why not use a `unique_ptr`?

niwhsa9 (changes requested):

jetson/nav/recovery.cpp
```diff
@@ -0,0 +1,206 @@
...
+    return seconds;
+}// getDuration()
+
+// returns the point the rover should be follwing for the recovery path
+RecoveryPoint Recovery::getPointToFollow() {
+    return currentRecoverypoint;
+}// getPointToFollow()
+
+// generate recovery path J turn
+void Recovery::makeRecoveryPath(const std::shared_ptr<Rover>& rover) {
```
niwhsa9, line 77: I'm honestly not a fan of passing smart pointers in function signatures, even by reference. In any modern C++ codebase it should be fairly clear that raw pointer handoffs do not signify change in ownership. Not a huge deal though.

niwhsa9: or even better, const reference

jetson/nav/search/searchStateMachine.cpp
```diff
@@ -3,9 +3,12 @@
 #include "utilities.hpp"
 #include "stateMachine.hpp"
 #include "searchFromPathFile.hpp"
+#include "stateMachine.hpp"
```
rbridges12, line 6: aren't all these added includes duplicates?

John-Trager: oops missed that I'll fix

jetson/nav/stateMachine.hpp
```diff
@@ -44,6 +45,8 @@ class StateMachine : public std::enable_shared_from_this<StateMachine> {
 
     double getDtSeconds();
 
+    std::unique_ptr<Recovery>& getRecovery();
```
qhdwight, line 48: Generally you never want to have a reference as an instance variable. This code is also undermining the point of a `unique_ptr` (the point is to have a single owner to a heap allocated variable) It helps to watch a couple videos such as https://www.youtube.com/watch?v=UOB7-B2MfwA to understand smart pointers in general.

As far as the references goes, think about it this way. A reference to a unique ptr has no owning factor. So if that unique ptr goes out of scope, you are screwed, you have a reference to garbage memory. This is why you never want to have a reference as a field unless you 100% know it will live past the lifetime of whatever instance has a reference to it. In fact I would argue you never want to do it as it is bad design and prone to bugs.

jetson/nav/stateMachine.hpp
```diff
@@ -11,10 +11,11 @@
 #include "obstacle_avoidance/simpleAvoidance.hpp"
 #include "rover.hpp"
 #include "search/searchStateMachine.hpp"
-
+#include "recovery.hpp"
 
 using namespace rover_msgs;
 using namespace std::chrono_literals;
+using duration =  std::chrono::duration<int>;
```
qhdwight, line 18: also generally people use `long long` as the `Rep` template parameter as it has higher precision. I think also clocks use nanoseconds under the hood so those can rack up fast and may not be held well by an `int`

niwhsa9: Would also recommend using steady clock for this since you are measuring deltas and dont care about absolute time. Steady clock is guaranteed never to go backwards..

In this particular case the time sync is not a huge concern since the granularity you need is on the order of seconds, but it's good practice.

jetson/nav/stateMachine.hpp
John-Trager: ok thank that makes sense, I should have looked up how to use smart pointers first. In my case the recovery object does need to be shared/used between the stateMachine and the searchStatemachine so should it be a `shared_ptr`? It really only needs to be owned by the stateMachine so I get why it doesn't need to be owned by the searchStateMachine so is there a better way to achieve this and still have it be used by both?

niwhsa9: Like you said, the owernship doesn't need to be shared. Since search state machine has a pointer to the regular state machine, you can use the recovery object as needed through that without assuming ownership of the unique ptr (don't perform moves). It is perfectly acceptable to use the raw pointer for non-owning functions and logic that just work with the Recovery.

jetson/nav/recovery.cpp
niwhsa9: just to clarify, I meant underlying object as in the object that the smart ptr contains.

jetson/nav/rover.cpp
```diff
@@ -112,6 +77,51 @@ bool Rover::drive(double distance, double bearing, double threshold, double dt)
     return false;
 }// drive()
 
+bool Rover::driveBackwards(const Odometry& destination, double stopDistance, double dt) {
+    double distance = estimateDistance(mOdometry, destination);
+    double bearing = estimateBearing(mOdometry, destination);
+    return driveBackwards(distance, bearing, stopDistance, dt);
+}// drive()
+
+bool Rover::driveBackwards(double distance, double bearing, double threshold, double dt) {
```
niwhsa9, line 86: This also seems mostly like a copy/pasta of the regular drive. Let's reduce duplication here too

jetson/nav/recovery.cpp
John-Trager: this should be resolved now

jetson/nav/stateMachine.hpp
John-Trager: this should also be fixed now as I just pass the underlying pointer rather than the ref of the unique_ptr

jetson/nav/stateMachine.hpp
John-Trager: it has now been changed to steady clock and long long as Rep

qhdwight: Nice, the `using` aliases are good too

niwhsa9: lgtm

PR #1156, opened 2022-08-01
[esw] Add auger and science arm support for CIRC 

tabiosg:


tabiosg:
Code is tested (except that I'm not sure if inversion is correct for raising/lowering the auger since I can't really tell since the end effector isn't fully functional right now).

jnnanni (approved):
Changes look good teleop-wise, assuming the inversion is right for the augur.

PR #1159, opened 2022-08-05
[teleop] Added Map to SA GUI, changed offline map format to png

jnnanni:
image
Map was changed to png because I finally figured out how to get map data directly from google satellite view. AWS Bucket has been updated with google maps data for CIRC. 

Someone with a working vagrant/jarvis needs to test this with the lcm bridge running and make certain that the map works as intended, I was unable to get jarvis working but was able to build the GUI using webpack. This meant I was able to run the GUI but no LCM bridge.

### DIFF
--- a/base_station/download_map/src/__main__.py
+++ b/base_station/download_map/src/__main__.py
@@ -3,7 +3,8 @@ import os
 import zipfile
 import json
 
-
+#If System has issues with number of file watcher when using GUI with downloaded map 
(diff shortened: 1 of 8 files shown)


tabiosg (approved):
It worked on one of the laptops! I'll also do it on the other ones for competition but it looks good to me. I'm pretty sure it covers a wide enough area and captures the competition area. I can go confirm with the others, but yeah looks good. Nice work!

rbridges12 (approved):

